PR #60, opened 2016-07-13
Fix setup missing `pybind11`

mihaic:


xoltar:
👍

mihaic:
It turns out it's not that simple. Even adding `pybind11` to `setup_requires` does not fix the problem:
```
brainiak/factor_analysis/tfa_extension.cpp:24:31: fatal error: pybind11/pybind11.h: No such file or directory
   #include <pybind11/pybind11.h>
```
Let me investigate further.

mihaic:
There is a difference in the behavior of `pip install` based on the package specification. When installing form Git, it installs the dependencies before building the wheel. When installing from PyPI, it first tries to build the wheel, and only then it installs the dependencies. Here are relevant excerpts:

`pip3 install -v git+https://github.com/IntelPNI/brainiak@v0.1`
```
  Using cached pybind11-1.8.1-py2.py3-none-any.whl
  Downloading from URL https://pypi.python.org/packages/e8/bb/cb21c4b8da28d94df193bd83fe399e13dd38b7eba6fed52546d4fb1c2908/pybind11-1.8.1-py2.py3-none-any.whl#md5=468347b8f3c1114c6cb23df6b4186bbb (from https://pypi.python.org/simple/pybind11/)
Installing collected packages: pybind11, brainiak

  Running setup.py install for brainiak ...
```

`pip install -v brainiak`
```
  Using cached pybind11-1.8.1-py2.py3-none-any.whl
  Downloading from URL https://pypi.python.org/packages/e8/bb/cb21c4b8da28d94df193bd83fe399e13dd38b7eba6fed52546d4fb1c2908/pybind11-1.8.1-py2.py3-none-any.whl#md5=468347b8f3c1114c6cb23df6b4186bbb (from https://pypi.python.org/simple/pybind11/)
Building wheels for collected packages: brainiak
  Running setup.py bdist_wheel for brainiak ...

[verbose output showing it fails because pybind11 is not installed]

Failed to build brainiak
Installing collected packages: pybind11, brainiak

  Running setup.py install for brainiak ...
```

mihaic:
This PR is wrong because it provides the wrong include path to the compiler (`setup_requires` does not install packages for usage after setup is finished). Closing and opening an issue instead to try to figure out the right solution.

Issue #65, opened 2016-07-14
Add a docstrings convention checker


mihaic:
E.g., https://github.com/PyCQA/flake8-docstrings

PR #72, opened 2016-07-19
Added Event Segmentation class, along with example and tests

cbaldassano:
Added event segmentation based on a hidden markov model. The paper describing this method is currently being written and will be published as a preprint within the next couple months.

### DIFF
--- a/brainiak/eventseg/__init__.py
+++ b/brainiak/eventseg/__init__.py
@@ -1,0 +1,14 @@
+#  Copyright 2016 Princeton University
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Event segmentation of continuous data + event transfer between datasets."""
--- a/brainiak/eventseg/event.py
+++ b/brainiak/eventseg/event.py
@@ -1,0 +1,369 @@
+#  Copyright 2016 Princeton University
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Event segmentation using a Hidden Markov Model
+
+Given an ROI timeseries, this class uses an annealed fitting procedure to
+segment the timeseries into events with stable activity patterns. After
+learning the signature activity pattern of each event, the model can then be
+applied to other datasets to identify a corresponding sequence of events.
+"""
+
+# Authors: Chris Baldassano and Cătălin Iordan (Princeton University)
+
+import numpy as np
+from scipy import stats
+import logging
+import copy
+from sklearn.base import BaseEstimator
+from sklearn.utils.validation import check_is_fitted, check_array
+
+logger = logging.getLogger(__name__)
+
+__all__ = [
+    "EventSegment",
+]
+
+
+class EventSegment(BaseEstimator):
+    """Class for event segmentation of continuous fMRI data
+
+    Parameters
+    ----------
+    n_events: int
+        Number of segments to learn
+
+    step_var: Callable[[int], float] : default 4 * (0.98 ** (step - 1))
+        The Gaussian variance to use during fitting, as a function of the
+        number of steps. Should decrease slowly over time.
+
+    n_iter: int : default 500
+        Maximum number of steps to run during fitting
+
+    Attributes
+    ----------
+    p_start, p_end: length n_events+1 ndarray
+        initial and final prior distributions over events
+
+    P: n_events+1 by n_events+1 ndarray
+        HMM transition matrix
+
+    ll_ : ndarray with length = number of training datasets
+        Log-likelihood for training datasets over the course of training
+
+    segments_:  list of (time by event) ndarrays
+        Learned (soft) segmentation for training datasets
+
+    event_var_ : float
+        Gaussian variance at the end of learning
+
+    event_pat_ : voxel by event ndarray
+        Learned mean patterns for each event
+    """
+
+    def _default_var_schedule(step):
+        return 4 * (0.98 ** (step - 1))
+
+    def __init__(self, n_events=2,
+                 step_var=_default_var_schedule,
+                 n_iter=500):
+        self.n_events = n_events
+        self.step_var = step_var
+        self.n_iter = n_iter
+
+    def fit(self, X, y=None):
+        """Learn a segmentation on training data
+
+        Fits event patterns and a segmentation to training data. After
+        running this function, the learned event patterns can be used to
+        segment other datasets using find_events
+
+        Parameters
+        ----------
+        X: time by voxel ndarray, or a list of such ndarrays
+            fMRI data to be segmented. If a list is given, then all datasets
+            are segmented simultaneously with the same event patterns
+
+        y: not used (added to comply with BaseEstimator definition)
+
+        Returns
+        -------
+        self: the EventSegment object
+        """
+
+        X = copy.deepcopy(X)
+        if type(X) is not list:
+            X = check_array(X)
+            X = [X]
+
+        n_train = len(X)
+        for i in range(n_train):
+            X[i] = X[i].T
+
+        n_dim = X[0].shape[0]
+        for i in range(n_train):
+            assert (X[i].shape[0] == n_dim)
+
+        # Double-check that data is z-scored in time
+        for i in range(n_train):
+            X[i] = stats.zscore(X[i], axis=1, ddof=1)
+
+        # Set up transition matrix, with final sink state
+        # For transition matrix of this form, the transition probability has
+        # no impact on the final solution, since all valid paths must take
+        # the same number of transitions
+        self.p_start = np.zeros((1, self.n_events + 1))
+        self.p_start[0, 0] = 1
+        self.P = np.vstack((np.hstack((
+            0.5 * np.diag(np.ones(self.n_events))
+            + 0.5 * np.diag(np.ones(self.n_events - 1), 1),
+            np.append(np.zeros((self.n_events - 1, 1)), [[0.5]], axis=0))),
+            np.append(np.zeros((1, self.n_events)), [[1]], axis=1)))
+        self.p_end = np.zeros((1, self.n_events + 1))
+        self.p_end[0, -2] = 1
+
+        self.classes_ = np.arange(self.n_events)
+
+        # Initialize variables for fitting
+        log_gamma = []
+        for i in range(n_train):
+            log_gamma.append(np.zeros((X[i].shape[1],
+                                       self.n_events)))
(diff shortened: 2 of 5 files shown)


buildbot-princeton:

buildbot-princeton:

mihaic:
Add to whitelist.

brainiak/event_seg/event.py
```diff
@@ -0,0 +1,312 @@
...
+logger = logging.getLogger(__name__)
+
+__all__ = [
+    "EventSegment",
+]
+
+
+class EventSegment:
+    """Class for learning and representing an event segmentation of
+    continuous fMRI data
```
mihaic: Please write a one-line summary (+ additional paragraphs for details, if needed):
https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

We have issue #65 open for adding this check to `pr-check`.

brainiak/event_seg/event.py
```diff
@@ -0,0 +1,312 @@
...
+
+    Attributes
+    ----------
+    n_events: int
+        Number of segments to learn
+
+    step_var: lambda : default lambda step: 4 * (0.98 ** (step - 1))
+        The Gaussian variance to use during fitting, as a function of the
+        number of steps. Should decrease slowly over time.
```
mihaic: As your description says, the type is `function`, not `lambda`.

brainiak/event_seg/event.py
```diff
@@ -0,0 +1,312 @@
...
+
+    max_steps: int : default 500
+        Maximum number of steps to run during fitting
+
+
+    p_start, p_end: row vectors
+        initial and final prior distributions over events
+
+    P: array
```
mihaic: Please add the shape for all arrays. Also, you should write `numpy.ndarray` or simply `ndarray` (unless you are using `array.array`).

brainiak/event_seg/event.py
```diff
@@ -0,0 +1,312 @@
...
+                 max_steps=500):
+        self.n_events = n_events
+        self.step_var = step_var
+        self.max_steps = max_steps
+
+        # Set up transition matrix, with final sink state
+        self.p_start = np.zeros((1, n_events + 1))
+        self.p_start[0, 0] = 1
+        self.P = np.vstack((np.hstack((0.5 * np.diag(np.ones(n_events)) +
+                                       0.5 * np.diag(np.ones(n_events - 1), 1),
```
mihaic: Binary operators should be at the beginning of the line, not at the end:
https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

brainiak/event_seg/event.py
```diff
@@ -0,0 +1,312 @@
...
+                            np.append(np.zeros((1, n_events)), [[1]], axis=1)))
+        self.p_end = np.zeros((1, n_events + 1))
+        self.p_end[0, -2] = 1
+
+        self.ll = None
+        self.segments = None
+        self.event_var = None
+        self.event_pat = None
+
+        return
```
mihaic: No need for empty `return`s.

tests/event_seg/test_event.py
```diff
@@ -0,0 +1,29 @@
...
+        "Segmentation from learn_events not correctly normalized"
+
+    T2 = 15
+    sample_data2 = np.random.rand(V, T2)
+    test_segments, test_ll = es.find_events(sample_data2)
+
+    assert test_segments.shape == (T2, K), "Segmentation from find_events " \
+                                           "has incorrect shape"
+    assert np.isclose(np.sum(test_segments, axis=1), np.ones(T2)).all(), \
+        "Segmentation from find_events not correctly normalized"
```
mihaic: You should add a simple manually-constructed example with well-understood results and test that you get those results.

mihaic:
It seems like you are close to implementing the scikit-learn Estimator and Predictor interfaces:
http://scikit-learn.org/stable/developers/contributing.html#apis-of-scikit-learn-objects

You should rename your `learn_events` method to `fit` and your `find_events` method to `predict`. Furthermore, you should add a trailing underscore to all data-dependent attributes set after fitting the model and not set them (even to `None`) outside `fit`:
http://scikit-learn.org/stable/developers/contributing.html#estimated-attributes

mihaic:
Please remove the underscore from the subpackage name, as we did with PR #73.

examples/event_seg/simulated_data.py
```diff
@@ -0,0 +1,77 @@
...
+    est_events2 = np.argmax(gamma, axis=1)
+    bound_match[1, test_i] = 1 - np.sum(abs(np.diff(event_labels2) -
+                                            np.diff(est_events2))) / (2 * K)
+
+# Across the testing datasets, print how well we identify the true event
+# boundaries and the log-likehoods in real vs. shuffled data
+print("Boundary match: {:.2} (null: {:.2})".format(
+    np.mean(bound_match[0, :]), np.mean(bound_match[1, :])))
+print("Log-likelihood: {:.3} (null: {:.3})".format(
+    np.mean(LL[0, :]), np.mean(LL[1, :])))
```
mihaic: Could you add a simple plot? We like colorful stuff. It's really a question, it's not mandatory.

brainiak/event_seg/event.py
mihaic: I was wrong about that. The type should be `Callable[[t1, t2, ..., tn], tr]`:
https://www.python.org/dev/peps/pep-0483/#fundamental-building-blocks

In case you are wondering, the requirement to document types like this in BrainIAK is documented (see second bullet):
https://github.com/IntelPNI/brainiak/blob/master/CONTRIBUTING.rst#standards

mihaic:
@cbaldassano Could you please name someone who is knowledgeable about your project, either from PNI or Intel?

cbaldassano:
I've made all the requested changes! @lcnature agreed last week to help review this code - Mingbo, I will send you the preprint describing the method to help you understand what it is supposed to be doing.

requirements-dev.txt
```diff
@@ -6,3 +6,5 @@ pytest-cython
 restructuredtext-lint
 sphinx
 sphinx_rtd_theme
+nose
```
mihaic: Please do not add an extra test framework. Please use `pytest`.

mihaic: Oh, I see. We can include `nose` in this case, if it actually works.

The lesson I learned is that comments for a demo during a meeting are not high quality. :)

cbaldassano: That's fine - if the sklearn tests weren't causing so many issues I would argue for including nose, but since I don't want to spend a ton of time debugging Python 3.4 pickling errors I took them out for now. I can say that the sklearn interface tests were passing at least in 3.5, so the current code meets all the primary BaseEstimator requirements.

mihaic:
@cbaldassano Have you sent Mingbo the preprint? Actually, are you considering publishing it on Arxiv or a similar place? If you are you could cite it in a docstring.

cbaldassano:
Yes, Mingbo has the preprint and said he will review the code soon. We'll be posting it on bioRxiv in a month or so, after we make some revisions, so maybe I'll submit a tiny pull request at that time to add the reference in the docstring.

brainiak/eventseg/event.py
```diff
@@ -0,0 +1,356 @@
...
+        -------
+        self: the EventSegment object
+        """
+
+        if type(X) is not list:
+            X = check_array(X)
+            X = [X]
+
+        n_train = len(X)
+        for i in range(n_train):
```
lcnature: I wonder if this can be avoided, e.g., by making copy of X and transpose the copy?
Because X is passed as reference, this line and line 118 which modify X will modify the data X outside the scope of this function, which might be unexpected by the user. For example, if a user wants to fit the data for multiple times, each time trying a different choice of K, (s)he might find the data has been modified after each time of fitting.

brainiak/eventseg/event.py
```diff
@@ -0,0 +1,356 @@
...
+
+        # Double-check that data is z-scored in time
+        for i in range(n_train):
+            X[i] = stats.zscore(X[i], axis=1, ddof=1)
+
+        # Set up transition matrix, with final sink state
+        self.p_start = np.zeros((1, self.n_events + 1))
+        self.p_start[0, 0] = 1
+        self.P = np.vstack((np.hstack((
+            0.5 * np.diag(np.ones(self.n_events))
```
lcnature: There is no complaint in this comment. But I am curious whether the choice of transition probability in self.P influences the result.

brainiak/eventseg/event.py
```diff
@@ -0,0 +1,356 @@
...
+            Log probability of each timepoint under each event Gaussian
+        """
+
+        n_vox = data.shape[0]
+        t = data.shape[1]
+
+        # z-score both data and mean patterns in space, so that Gaussians
+        # are measuring Pearson correlations and are insensitive to overall
+        # activity changes
+        data = stats.zscore(data, axis=0, ddof=1)
```
lcnature: I suspect that this line might run the same risk: it will z-score the original data passed to the fit() function, unless the original data is passed as a copy.

brainiak/eventseg/event.py
```diff
@@ -0,0 +1,356 @@
...
+        -------
+        segments : time by event ndarray
+            The resulting soft segmentation. segments[t,e] = probability
+            that timepoint t is in event e
+
+        test_ll : float
+            Log-likelihood of model fit
+        """
+
+        testing_data = testing_data.T
```
lcnature: To avoid modifying data outside the function, perhaps you can replace the testing_data in line 333 with testing_data.T

brainiak/eventseg/event.py
```diff
@@ -0,0 +1,356 @@
...
+                log_alpha[0, :] = self._log(self.p_start) + logprob[0, :]
+            else:
+                log_alpha[t, :] = self._log(np.exp(log_alpha[t - 1, :])
+                                            .dot(self.P)) + logprob[t, :]
+
+            log_scale[t] = np.logaddexp.reduce(log_alpha[t, :])
+            log_alpha[t] -= log_scale[t]
+
+        # Backward pass
+        log_beta[-1, :] = self._log(self.p_end) - log_scale[-1]
```
lcnature: log_scale was calculated based on log_alpha in line 255, but not recalculated here based on log_beta, but it . I wonder if you need to do something similar as in line 255-256 but for log_beta. But since line 270 directly uses the log_scale calculated in line 255, perhaps you actually did not want to include anything from the backward pass in the ll? Sorry I do not know much of the forward-backward algorithm.

brainiak/eventseg/event.py
cbaldassano: Very good point - I'll only modify copies.

brainiak/eventseg/event.py
cbaldassano: Yes, will fix.

brainiak/eventseg/event.py
cbaldassano: Yes, will fix.

brainiak/eventseg/event.py
cbaldassano: This is actually correct, though I agree it looks asymmetrical. I'll add a link to the wikipedia page which explains what's going on with the log_scale parameter (called c_i on wiki) https://en.wikipedia.org/wiki/Forward%E2%80%93backward_algorithm

mihaic:
@lcnature Are you happy with this PR?

brainiak/eventseg/__init__.py
```diff
@@ -0,0 +1,14 @@
+#  Copyright 2016 Intel Corporation
```
mihaic: This should be Princeton or your name.

brainiak/eventseg/event.py
```diff
@@ -0,0 +1,363 @@
...
+logger = logging.getLogger(__name__)
+
+__all__ = [
+    "EventSegment",
+]
+
+
+class EventSegment(BaseEstimator):
+    """Class for learning and representing an event segmentation of
+    continuous fMRI data
```
mihaic: Docstrings should start with a one-line summary:
https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

lcnature:
@mihaic @cbaldassano 
Sorry for the delay. I am happy with all the changes :+1:

mihaic:
Thank you very much for your contribution, @cbaldassano. And thanks for the review, @lcnature.

PR #87, opened 2016-07-29
hyperparamopts: Fix Cython missing after install

mihaic:


TuKo:
Is that the only requirements-dev.txt file cython should appear?

TuKo:
:+1:

PR #88, opened 2016-08-03
add image for travis in readme

cameronphchen:
add image of travis-ci to readme


mihaic:
I will add it, but it is superfluous because master cannot break with our workflow (really).

PR #89, opened 2016-08-03
Update version to 0.2

mihaic:


xoltar:
Should we add a CHANGES.md to describe what is new in each release?

mihaic:
I think it is too error prone and time consuming to keep a manually edited file up-to-date. I saw a couple of generators that extract issue and PR information from GitHub, but I don't have time to evaluate them now. I propose we make an issue and address it later.

xoltar:
OK, sounds good. 
👍

PR #96, opened 2016-08-17
add example code to run fcma executable file built from the original FCMA toolbox

yidawang:
It assumes FCMA has been installed. The input arguments are parsed by the Python code and passed to the FCMA executable via subprocess.


yidawang:
I will re-open this when I get back on it

Issue #42, opened 2016-06-28
dev: Show source code line numbers in doc build errors


mihaic:
Docstring line numbers are shown currently, which makes it difficult to see to find the error in the source code.

PR #98, opened 2016-09-07
Add Bayesian RSA to the package (reprsimil)

lcnature:
Bayesian RSA stands for Bayesian representational similarity analysis.
The code should be functional.
Planned updates for the upcoming weeks:
(1) add an example as a python notebook
(2) upload supplementary file including the derivation of the gradient of log likelihood
(3) capability of modeling the spatial correlation of noise


buildbot-princeton:

buildbot-princeton:

mihaic:
Add to whitelist.

mihaic:
Thank you for the PR, @lcnature.

Could you please recommend someone who is familiar with your work and can assess the neuroscience validity of the code?

Also, is `pr-check` passing for you? I tried on my computer and `run-checks` is failing just like on Jenkins because of trailing whitespace:
```
brainiak/reprsimil/brsa.py:23:31: W291 trailing whitespace
brainiak/reprsimil/brsa.py:123:14: W291 trailing whitespace
```

lcnature:
Sorry. I made some changes to the docstring after it passed the last time.
The trailing whitespace was fixed but now I have errors about unexpected indentation in docstring, which I cannot find why.

mihaic:
That's perfectly OK. You can make as many commits you want, there is no issue with having test failures. :)

The line number in the doc building errors are very misleading (see issue #42).
 
You need an empty line before "Parameters".

lcnature:
Thanks @mshvartsman who has kindly agreed to review. I will let you know after I upload more documentation of the maths.

lcnature:
@mihaic How can I know why the check with macos failed?

yidawang:
@lcnature Within Princeton network, you can click the Details to see. It looks like it is an import error related to FCMA, which is weird because FCMA is not yet merged. @mihaic may have some idea about this?

lcnature:
@yidawang In fact I do not have read permission -- access denied even after logging in.

yidawang:
@lcnature I don't even know there are permission settings out there. In the "console output" I can see the code tries to import brainiak.fcma.cython_blas which is not supposed to be in the current codebase.

mihaic:
Retest this please.

mihaic:
The error was caused by leftovers from previous PR builds, which were preventing the documentation from building. So not related to the new code. I added a clean step to the builder (I am sure I had added it before and it magically disappeared...) to prevent this from happening again.

Unfortunately, the Jenkins results are only visible to people with Jenkins accounts. I will talk to the admin to see if we can make the results public.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+""" Bayesian Representational Similarity Analysis (BRSA)
+
+    This implementation is based on the following publications:
+
+ .. [Cai2016] "Unbiased Bayesian estimation of
+    neural representational similarity structure",
+    M.B. Cai, N. Schuck, J. Pillow, Y. Niv,
+    Neural Information Processing System 29, 2016.
```
mshvartsman: small typo: Neural Information Processing System**s**

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+
+    Notes
+    -----
+    The current version assumes noise is independent across voxels.
+    Real data typically has spatial correlation in noise.
+    This assumption might still introduce some bias in the result.
+    Spatial correlation will be included in a future version.
+    """
+
+    def __init__(
```
mshvartsman: Looks like there are some parameters here (intern_smooth_range, space_smooth_range, etc) that are not documented up in the docstring. If I understand the standard correctly, all exposed parameters (e.g. arguments to __init___) should be documented.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+        return
+
+    def fit(self, X, design, scan_onsets=None, coords=None,
+            inten=None, inten_weight=None):
+        """Compute the Bayesian RSA
+
+        Parameters
+        ----------
+        X: 2-D numpy array, shape=[time_points, voxels]
+            In the equation, this is actually Y.
```
mshvartsman: This is confusing because of the shift from generative to descriptive perspective -- but I am not sure what the best solution is. At the very least, the derivation that ships with brainiak should match notation with the code and also show the relationship to the NIPS paper notation. But maybe there is a better option, maybe someone else can think of it?

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+                    inten_weight = self.inten_weight
+                assert inten_weight >= 0 and inten_weight <= 1,\
+                    'The weight of intensity kernel is '\
+                    'out of the range of (0,1).'\
+                    'I suggest using a small value, such as 0.1.'
+            else:
+                inten_weight = 0.0
+
+        # Run Bayesian RSA
+        if not self.GP_space:
```
mshvartsman: I think right now if I pass in onsets without asking for the GP prior then they get silently ignored. Would it be better to check above for incompatible combinations of parameters like this, and then just let `scan_onsets` get passed through here? Just having some trouble following the logic -- another thing that might help is just commenting on what the different dispatchers are (I think it's BRSA without GP spatial prior for the first one, then BRSA without GP prior on intensity, then BRSA with both).

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+            self.U_, self.L_, self.nSNR_, self.sigma_, self.rho_, \
+                self.lGPspace_, self.bGP_, self.lGPinten_ = \
+                self._fit_RSA_UV(design, X, scan_onsets=scan_onsets,
+                                 coords=coords, inten=inten,
+                                 inten_weight=inten_weight)
+
+        if self.pad_DC:
+            self.U_ = self.U_[:-1, :-1]
+            self.L_ = self.L_[:-1, :self.rank]
+        sqrt_diag = np.diag(self.U_)**0.5
```
mshvartsman: Maybe more elegant: 
```
inv_sd = 1/np.sqrt(np.diag(self.U_))
self.C_ = self.U_ * np.outer(inv_sd, inv_sd)
```
(or even better yet, define `__cov2cor` elsewhere and call it here)

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+            # a new scan starts. For example, if n_T = 500, and
+            # scan_onsets = [0,100,200,400], this means that the time points
+            # of 0-99 are from the first scan, 100-199 are from the second,
+            # 200-399 are from the third and 400-499 are from the fourth
+            run_TRs = np.diff(np.append(scan_onsets, n_T))
+            logger.debug('I infer that the number of volumes'
+                         ' in each scan are: {}'.format(run_TRs))
+
+            D_ele = list(map(self._D_gen, run_TRs))
+            F_ele = list(map(self._F_gen, run_TRs))
```
mshvartsman: Do you need to turn these iterators to lists or can you just iterate on them directly?

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+            n_smooth = 0
+            dist2 = None
+            inten_diff2 = None
+            GP_inten = False
+            space_smooth_range = None
+            inten_smooth_range = None
+        return dist2, inten_diff2, space_smooth_range, inten_smooth_range,\
+            n_smooth
+
+    def _fit_RSA_UV(self, design, Y,
```
mshvartsman: `X` turns into `Y` on this call (from line 264/268/272). But `design` is still `design` and doesn't turn into `X` until the intermediate variables computed in `_prepare_data`. Assuming you're going to keep the scikit learn standard in the interface and shift to your notation internally, it is probably worth (a) making the switch in a single location, and (b) calling it out explicitly.

brainiak/reprsimil/brsa.py
lcnature: Thanks Mike! I will add them! Those are parameters added later and I forgot
to add documentation.

Mingbo Cai


On Thu, Sep 8, 2016 at 1:39 PM, Michael Shvartsman <notifications@github.com

brainiak/reprsimil/brsa.py
lcnature: Yeah, I also do not have a good idea of what to do.
@mihaic Basically the issue is in our model, we call X the design matrix (things we believe we know to some extent) and Y the data (which from our generative model is generated from X). But in scikit-learn, the input data is called X. So I have to make the explanation that what I write in the wrapper function fit() as X is actually Y in my major fitting code and the paper.

brainiak/reprsimil/brsa.py
lcnature: Thanks @mshvartsman . I am not understanding this. What will be ignored if onsets are passed but not asking for the GP prior? I will add the comments on each dispatcher. Good idea!

mshvartsman: Suppose I pass in onsets, but also do not ask for GP prior. Then as far as I can tell the model just ignores the fact that I pass in onsets? Maybe this is fine behavior, though I would expect at least a warning or something if I pass in irrelevant arguments.

lcnature: Ah, it does not seem to be. if self.GP_space is not True (also when it is None), then this line is executed:
self.U_, self.L_, self.nSNR_, self.sigma_, self.rho_ = \
                self._fit_RSA_UV(design, X, scan_onsets=scan_onsets)
Then scan_onsets is passed to self._fit_RSA_UV which will do
XTY, XTDY, XTFY, YTY_diag, YTDY_diag, YTFY_diag, XTX, XTDX, XTFX = \
            self._prepare_data(design, Y, n_T, n_V, scan_onsets)
The scan onsets are considered in the inverse of the temporal covariance matrix of noise.

mshvartsman: Sorry, I don't know why I wrote `scan_onsets`, I see why my comment didn't make sense. I meant to make the point w.r.t. the other variables like `coords`. On line 197 you check I try to pass `GP_inten` without passing `GP_space`, but there's no similar checks for the other variables that only make sense conditional on others being defined. Or maybe I missed it.

brainiak/reprsimil/brsa.py
lcnature: Great idea! Will add the function to utils.py

brainiak/reprsimil/brsa.py
mihaic: The code should match the rest of BrainIAK and Scikit-learn, so the variable representing input data in the `fit` method should be called `X`. Furthermore, `y` is reserved for labels. Perhaps it is best to rename `Y` to `X` and use `design` throughout the code, even if it creates a mismatch with the paper.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+""" Bayesian Representational Similarity Analysis (BRSA)
+
+    This implementation is based on the following publications:
+
+ .. [Cai2016] "Unbiased Bayesian estimation of
+    neural representational similarity structure",
+    M.B. Cai, N. Schuck, J. Pillow, Y. Niv,
+    Neural Information Processing System 29, 2016.
+    A preprint is available at
+    http://biorxiv.org/content/early/2016/09/07/073932
```
mihaic: You should use the DOI of the preprint:
https://doi.org/10.1101/073932

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+        # calculate the square of difference between each voxel's location
+        # coorinates and image intensity.
+        if GP_space:
+            assert coords is not None, 'coordinate is not provided'
+            n_V = coords.shape[1]
+            dist2 = np.zeros([n_V, n_V])
+            # square of spatial distance between every two voxels
+            for dim in range(np.size(coords, 1)):
+                coord_tile = np.tile(coords[:, dim], [n_V, 1])
+                dist2 = dist2 + (coord_tile - coord_tile.T)**2
```
mshvartsman: `np.square` is probably better here.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+            * param0_fitU[n_l:n_l + n_V], axis=0))
+        param0_fitU[n_l:n_l + n_V] = \
+            np.tan(param0_fitU[n_l:n_l + n_V] * np.pi / 2)
+        # up to this point, the GP parameters are not set.
+
+        param0[0:n_l] = current_vec_U_chlsk_l_AR1
+        param0[n_l] = np.median(log_sigma2_est)
+        param0[n_l + 1] = np.median(param0_fitU[n_l:n_l + n_V])
+
+        res = scipy.optimize.minimize(
```
mshvartsman: Would be useful to expose this rather than hardcode the optimizer.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+
+        # (3) random initialization
+
+        current_vec_U_chlsk_l_AR1 = np.random.randn(n_l)
+
+        param0 = np.zeros(n_l + 2)
+        param0_fitU = np.zeros(n_l + n_V)
+        param0_fitV = np.zeros(n_V + n_smooth - 1)
+
+        param0_fitU[0:n_l] = current_vec_U_chlsk_l_AR1
```
mshvartsman: Here and elsewhere, would be good to have an array of indices rather than hardcode here -- makes it much harder to remember the organization of the parameters in the param vector.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+        # estimation of covariance, just in case
+        # the user provides data in which
+        # n_V is smaller than n_C
+
+        # (2) start from identity matrix
+
+        # current_vec_U_chlsk_l_AR1 = np.eye(n_C)[l_idx]
+
+        # (3) random initialization
+
```
mshvartsman: Should probably explain that you first fit a simplified model to produce an initial guess for the full model.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+        # this threshold, we stop fitting.
+
+        # We cannot start fitting smoothness parameter just after
+        # fitting the single parameter version. The reason is
+        # that we would have SNR equal in all voxels to start with.
+        # This will favor a very large smoothness length scale.
+        # Therefore, we first fit for some iterations
+        # without the smoothness parameter.
+
+        if GP_space:
```
mshvartsman: In general this should be multiple functions. Probably the best way is to have functions for estimating U and V in the full model, plus the constant-noise estimate, plus the no-GP estimate. Then the sequence of calls in the main code will make it very clear what the full iterative estimation looks like. Right now it is hard to follow.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+                current_logSNR2[0:n_V - 1] = res_fitV.x[0:n_V - 1]
+                current_logSNR2[-1] = - np.sum(current_logSNR2[0:n_V - 1])
+
+                norm_fitVchange = np.linalg.norm(res_fitV.x[:n_V - 1]
+                                                 - param0_fitV[:n_V - 1])
+                logger.debug('norm of parameter change after fitting V: '
+                             '{}'.format(norm_fitVchange))
+                logger.debug('E[log(SNR2)^2]:', np.mean(current_logSNR2**2))
+
+                # Debugging purpose. Commented out due to complexity limit.
```
mshvartsman: This is not the right way of solving the complexity limit problem :). Probably after refactoring this into multiple functions as suggested above, you will be able to uncomment this.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+                    options={'xtol': 10**-3, 'disp': self.verbose,
+                             'maxiter': 3})
+                current_vec_U_chlsk_l_AR1 = res_fitU.x[0:n_l]
+                current_a1 = res_fitU.x[n_l:n_l + n_V]
+                norm_fitUchange = np.linalg.norm(res_fitU.x - param0_fitU)
+                logger.debug('norm of parameter change after fitting U: '
+                             '{}'.format(norm_fitUchange))
+                param0_fitU = res_fitU.x.copy()
+
+                # if norm_fitVchange / np.sqrt(param0_fitV.size - n_smooth)\
```
mshvartsman: Uncommenting out this convergence check will probably speed things up by not overdoing the optimization?

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+                #         < tolx and norm_fitUchange\
+                #         / np.sqrt(param0_fitU.size) < tolx:
+                #     break
+
+            current_GP[0] = np.log(np.min(
```
mshvartsman: Consider doing multiple restarts with different initial length scales, since (per in person discussion now) this seems to make a difference.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+                dist2[np.tril_indices_from(dist2, k=-1)]))
+#             current_GP[0] = np.log(np.max(dist2)/4.0)
+            # Initialize it with the smallest distance.
+            # Alternatively, initialize with a large distance.
+            # In testing, the result does depend on the initialization.
+            # There can be two local minimum when the data does not
+            # constrain the spatial landscape of log(SNR2) enough
+            # I think this is because of the addition of epsilon.
+            if GP_inten:
+                current_GP[1] = np.log(np.maximum(
```
mshvartsman: Should explain this heuristic.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+                    np.percentile(inten_diff2[np.tril_indices_from(
+                        inten_diff2, k=-1)], 5), 0.5))
+
+            param0_fitV[n_V - 1:n_V - 1 + n_smooth] = current_GP
```
mshvartsman: Use index array here.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+            logger.debug('current GP[0]:{}'.format(current_GP[0]))
+
+        tolx = self.tolx
+        n_iter = self.n_iter
+        logger.debug('start real fitting')
```
mshvartsman: Would be useful to be more clear on what "real fitting" means.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+            # fit V
+
+            res_fitV = scipy.optimize.minimize(
+                self._loglike_y_AR1_diagV_fitV, param0_fitV, args=(
+                    XTX, XTDX, XTFX, YTY_diag, YTDY_diag, YTFY_diag, XTY, XTDY,
+                    XTFY, current_vec_U_chlsk_l_AR1, current_a1, l_idx,
+                    n_C, n_T, n_V, rank, GP_space, GP_inten, dist2,
+                    inten_diff2, space_smooth_range, inten_smooth_range,
+                    inten_weight),
+                method='BFGS', jac=True, tol=10**(-2 - 2 / n_iter * (it + 1)),
```
mshvartsman: Tighter tolerance with increasing iterations is good -- but it's hard to figure out what the decay schedule is. Would be good to comment (and also respect tolx if it is being passed in).

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+            current_logSNR2[0:n_V - 1] = res_fitV.x[0:n_V - 1]
+            current_logSNR2[n_V - 1] = -np.sum(current_logSNR2[0:n_V - 1])
+
+            fitVchange = res_fitV.x[:n_V - 1] - param0_fitV[:n_V - 1]
+            norm_fitVchange = np.linalg.norm(fitVchange)
+
+            param0_fitV = res_fitV.x.copy()
+            logger.debug('norm of parameter change after fitting V: '
+                         '{}'.format(norm_fitVchange))
+            logger.debug('E[log(SNR2)^2]: {}'.format(
```
mshvartsman: Clarify why this is very informative about whether there is signal in the data (and how to interpret).

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1118 @@
...
+            if np.max(np.abs(fitVchange)) < tolx and \
+                    np.max(np.abs(fitUchange)) < tolx:
+                break
+
+        Uest_chlsk_l_AR1_UV = np.zeros([n_C, rank])
+        Uest_chlsk_l_AR1_UV[l_idx] = current_vec_U_chlsk_l_AR1
+
+        est_cov_AR1_UV = np.dot(Uest_chlsk_l_AR1_UV, Uest_chlsk_l_AR1_UV.T)
+
+        opt_rho1_AR1_UV = 2 / np.pi * np.arctan(current_a1)
```
mshvartsman: Worth checking whether tan or scaled logit is a better transformation to unconstrained space (w.r.t. speed, quality of estimate, and the gradients). I thought logit was more standard for this (but maybe it's just stan).

brainiak/reprsimil/brsa.py
mshvartsman: Also: just ran into a similar problem elsewhere and I wonder if http://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.pdist.html will do what you want without reinventing the wheel?

mshvartsman:
By the way, I am commenting on every little detail I see, but Mingbo and I have been chatting in person and only some of them are really necessary to fix before merging, many others will go into a code cleanup issue for later revisions -- don't mean to hold this up for too long.

brainiak/reprsimil/brsa.py
lcnature: Awesome! The sqeclidean is what we wanted.

brainiak/reprsimil/brsa.py
lcnature: The new version looks more like what Mike suggested: in the wrapper, X and y are used the same as in scikit-learn. In other places, it follows the paper and common notation in fMRI literature

brainiak/reprsimil/brsa.py
lcnature: A warning is now added for this case. Thanks @mshvartsman

brainiak/reprsimil/brsa.py
lcnature: fixed!

brainiak/reprsimil/brsa.py
lcnature: Fixed!

mihaic:
@mshvartsman I appreciate the thoroughness of your review. If at any point you are satisfied with the PR, please say so. If at that point there are comments left to be addressed later, it would be great to turn them into issues after the merge.

mshvartsman:
I'll try to clarify what I think can be saved for fixing later.

brainiak/utils/utils.py
```diff
@@ -100,3 +102,128 @@ def fast_inv(a):
...
+    assert cov.ndim == 2, 'covariance matrix should be 2D array'
+    inv_sd = 1 / np.sqrt(np.diag(cov))
+    corr = cov * np.outer(inv_sd, inv_sd)
+    return corr
+
+
+class read_design:
+    """ A class which has the ability of reading in design matrix in .1D file,
+        generated by AFNI's 3dDeconvolve.
+        In future, we will make it read SPM's design matrix file as well.
```
mshvartsman: This is good to document in a new issue rather than in the docs (can do that after the merge though).

lcnature:
@mihaic @mshvartsman In fact I have not fix my codes according to all what Mike suggested yet.

brainiak/reprsimil/brsa.py
lcnature: These are now all changed to fix tolerance

brainiak/reprsimil/brsa.py
lcnature: Done :)

brainiak/reprsimil/brsa.py
lcnature: I just added a half-Cauchy prior on the standard deviation of the Gaussian process in addition to the length scale. This seems to have avoided the other convergence point (super huge length scale together with super huge standard deviation)

brainiak/reprsimil/brsa.py
lcnature: They are uncommented now

brainiak/reprsimil/brsa.py
lcnature: Thanks @mshvartsman This is done now.

brainiak/reprsimil/brsa.py
lcnature: Done

lcnature:
@mihaic @mshvartsman The current version should have improved a lot.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1316 @@
...
+
+        SNR2 = np.exp(log_SNR2)
+        # each element of SNR2 is the ratio of the diagonal element on V
+        # to the variance of the fresh noise in that voxel
+
+        # derivatives
+        deriv_L = np.zeros(np.shape(L))
+        deriv_a1 = np.zeros(np.shape(rho1))
+
+        YTAY = YTY_diag - rho1 * YTDY_diag + rho1**2 * YTFY_diag
```
mshvartsman: Would be really really good to reference these lines to expressions / subexpressions in the paper.

brainiak/reprsimil/brsa.py
```diff
@@ -0,0 +1,1316 @@
...
+            - np.sum(np.log(np.linalg.det(LAMBDA_i))) * 0.5 \
+            - n_T / 2.0
+        XTAXL = np.dot(XTAX, L)
+        # dimension: space*feature*rank
+        deriv_L = -np.einsum('ijk,ikl,i', XTAXL, LAMBDA, SNR2) - \
+            np.einsum('ijk,ik,il,i', XTAXL, YTAXL_LAMBDA, YTAXL_LAMBDA,
+                      SNR2**2 / sigma2) \
+            + np.dot(XTAY / sigma2 * SNR2, YTAXL_LAMBDA)
+        # dimension: feature*rank
+        tmp = -XTDX + 2 * rho1[:, np.newaxis, np.newaxis] * XTFX
```
mshvartsman: Any way to name `tmp` more informatively?

brainiak/utils/utils.py
```diff
@@ -100,3 +102,127 @@ def fast_inv(a):
...
+            self.design = np.zeros([0, 0])
+            self.n_col = 0
+            # number of columns (conditions) in the design matrix
+            self.column_types = np.ones(0)
+            self.n_basis = 0
+            self.n_stim = 0
+            self.n_orth = 0
+            self.StimLabels = []
+        else:
+            isAFNI = re.match(r'.+[.](1D|1d|txt)$', fname)
```
mshvartsman: Should probably be `os.path.splitext`.

brainiak/utils/utils.py
```diff
@@ -100,3 +102,127 @@ def fast_inv(a):
...
+    assert cov.ndim == 2, 'covariance matrix should be 2D array'
+    inv_sd = 1 / np.sqrt(np.diag(cov))
+    corr = cov * np.outer(inv_sd, inv_sd)
+    return corr
+
+
+class read_design:
+    """ A class which has the ability of reading in design matrix in .1D file,
+        generated by AFNI's 3dDeconvolve.
+    """
```
mshvartsman: Arguments not documented here.

mshvartsman: Also: worth checking if this and readAFNI below are duplicating nipype functionality.

brainiak/utils/utils.py
```diff
@@ -100,3 +102,127 @@ def fast_inv(a):
...
+        # Read design file written by AFNI
+        self.n_basis = 0
+        self.n_stim = 0
+        self.n_orth = 0
+        self.StimLabels = []
+        self.design = np.loadtxt(fname, ndmin=2)
+        with open(fname) as f:
+            all_text = f.read()
+
+        find_n_column = re.compile(
```
mshvartsman: Really basic error checking would be nice (just so that it doesn't fail silently if it doesn't find what it needs).

tests/reprsimil/test_brsa.py
```diff
@@ -0,0 +1,211 @@
...
+    u_b = brsa.U_[1:,1:]
+    u_i = ideal_cov[1:,1:]
+    p = scipy.stats.spearmanr(u_b[np.tril_indices_from(u_b,k=-1)],u_i[np.tril_indices_from(u_i,k=-1)])[1]
+    assert p < 0.01, "Fitted covariance matrix does not correlate with ideal covariance matrix!"
+    # check that the recovered SNR makes sense
+    p = scipy.stats.pearsonr(brsa.nSNR_,snr)[1]
+    assert p < 0.01, "Fitted SNR does not correlate with simualted SNR!"
+    assert np.isclose(np.mean(np.log(brsa.nSNR_)),0), "nSNR_ not normalized!"
+
+    assert np.abs(brsa.bGP_ - tau) / tau < 0.3, "standard deviation of GP deviates too much"
```
mshvartsman: These should probably be (eventually) replaced with more theoretically motivated bounds. Depending on how easy this is, you could do it now or later.

tests/reprsimil/test_brsa.py
```diff
@@ -0,0 +1,211 @@
...
+    assert np.abs(brsa.lGPspace_ - smooth_width) / smooth_width < 0.7,\
+        "spatial length scale of GP deviates too much"
+    assert np.abs(brsa.lGPinten_ - inten_kernel) / inten_kernel < 0.7,\
+        "intensity length scale of GP deviates too much"
+    
+    # test if gradient is correct
+    XTY,XTDY,XTFY,YTY_diag, YTDY_diag, YTFY_diag, XTX, XTDX, XTFX = brsa._prepare_data(design.design_used,Y,n_T,n_V,scan_onsets)
+    n_l = n_C*(n_C+1)/2
```
mshvartsman: Probably should do this using something like http://numdifftools.readthedocs.io/en/latest/ rather than simple first order (will also allow to shrink tolerance). Other alternative is to compare against a symbolic diff tool like theano but that's more work so maybe later.

brainiak/utils/utils.py
lcnature: I think AFNI still does not provide easy python tool for reading relevant information from 1D files. Basically the support for python is very limited: https://afni.nimh.nih.gov/afni/community/board/read.php?1,82973,82973#msg-82973

brainiak/utils/utils.py
lcnature: Actually I realized that there is a checking a few lines below, for whether the number of conditions read from the header matches the number of columns.

tests/reprsimil/test_brsa.py
lcnature: I prefer to leave this for future issue :)

tests/reprsimil/test_brsa.py
lcnature: @mshvartsman It seems that this package needs to be installed additionally. Since this is only used in the testing code, I am not sure if it is worth adding it to the requirement list  -- somehow I remember @mihaic suggested not changing the requirement list but I am not sure. I changed the perturbation to a much smaller value and thus the tolerance was shrunk to a lower value.

brainiak/utils/utils.py
lcnature: Changed.

brainiak/reprsimil/brsa.py
lcnature: I renamed it as dXTAX_drho1 now - the derivative of the product X'AX wrt to the AR(1) coefficient, where X'AX is explained somewhere above. A/sigma2 is the inverse covariance matrix of noise in each voxel.

tests/reprsimil/test_brsa.py
mihaic: We prefer adding as few dependencies as possible, but if some functionality exists somewhere else, we do prefer adding a dependency to duplicating the functionality. This is especially the case if the dependency is only for testing and not for using BrainIAK. I looked at `numdifftools` and I see it only has one additional dependency (`algopy`), which has no additional dependencies. Feel free to add `numdifftools` to `requirements-dev.txt` and use it for testing.

lcnature:
@mihaic I cannot tell what causes a failure in Travid CI test:
https://travis-ci.org/IntelPNI/brainiak/jobs/161177341

mihaic:
@lcnature No clue. The error was only on MacOS. The tests passed on MacOS on Jenkins, so I restarted the Travis test.

mihaic:
@lcnature More evidence that the best approach for solving computer errors is restarting. :)

lcnature:
@mshvartsman

tests/reprsimil/test_brsa.py
lcnature: The test now uses numdifftools. The testing time is about 8 minutes, which used to be ~ 4 min. If the timing is fine with @mihaic , can you decide whether there is any other issue? @mshvartsman

mihaic: The test duration is quite long, but we can improve it later.

lcnature: I reduced the number of voxels to 200 in the test and it gets a bit shorter now.

brainiak/utils/utils.py
```diff
@@ -100,3 +103,164 @@ def fast_inv(a):
...
+        if self.include_pols:
+            self.cols_used = np.sort(np.append(
+                self.cols_used, np.where(self.column_types == -1)[0]))
+        self.design_used = self.design[:, self.cols_used]
+        if not self.include_pols:
+            # baseline is not included, then we add a column of all 1's
+            self.design_used = np.insert(self.design_used, 0, 1, axis=1)
+        self.n_TR = np.size(self.design_used, axis=0)
+
+    def readAFNI(self, fname):
```
mihaic: Method names should be lowercase, with optional underscores for readability:
https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

brainiak/utils/utils.py
```diff
@@ -100,3 +103,164 @@ def fast_inv(a):
...
+
+
+    """
+    assert cov.ndim == 2, 'covariance matrix should be 2D array'
+    inv_sd = 1 / np.sqrt(np.diag(cov))
+    corr = cov * inv_sd[None, :] * inv_sd[:, None]
+    return corr
+
+
+class read_design:
```
mihaic: Class names should be use the CapWords capitalization convention:
https://www.python.org/dev/peps/pep-0008/#class-names
While for the code that closely follows the formulas in the paper we can disregard this rule, please follow it here.

lcnature:
This time travis-CI throws some details for its failure. It looks like some problem with pytest iteslf.

brainiak/utils/utils.py
lcnature: Thanks. Done

brainiak/utils/utils.py
lcnature: Done.

mihaic:
@lcnature Please update the branch so I can merge it.

lcnature:
@mihaic Thanks Mihai, I updated and travis-CI failed again...

mihaic:
@lcnature Thank you very much for your contribution and for addressing the issues that came up during the review (thanks for pointing them out, @mshvartsman).
We will be looking into the Travis failure.

PR #99, opened 2016-09-12
use MPI for tests

xiazhu:
This is to update  TFA/HTFA tests based on MPI requirement from FCMA code.

Thanks in advance for your review.

### DIFF
--- a/tests/factoranalysis/test_htfa.py
+++ b/tests/factoranalysis/test_htfa.py
@@ -126,37 +126,39 @@ def test_can_run():
         if idx % size == rank:
             my_R.append(R[idx])
 
-    htfa = HTFA(
-        K,
-        n_subj=n_subj,
-        max_global_iter=max_global_iter,
-        max_local_iter=max_local_iter,
-        max_voxel=max_voxel,
-        max_tr=max_tr,
-        verbose=True)
-    assert htfa, "Invalid HTFA instance!"
+    for root in range(size):
+        htfa = HTFA(
+            K,
+            n_subj=n_subj,
+            max_global_iter=max_global_iter,
+            max_local_iter=max_local_iter,
+            max_voxel=max_voxel,
+            max_tr=max_tr,
+            verbose=True,
+            root=root)
+        assert htfa, "Invalid HTFA instance!"
 
-    X = []
-    for s in np.arange(n_subj):
-        X.append(np.random.rand(n_voxel, n_tr))
-    my_data = []
-    for idx in np.arange(n_subj):
-        if idx % size == rank:
-            my_data.append(X[idx])
+        X = []
+        for s in np.arange(n_subj):
+            X.append(np.random.rand(n_voxel, n_tr))
+        my_data = []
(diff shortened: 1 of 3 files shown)


tests/factoranalysis/test_htfa.py
```diff
@@ -126,37 +126,39 @@ def test_can_run():
...
-
-    else:
-        htfa.fit(my_data, R=my_R)
-        assert True, "worker successfully running HTFA"
-        print(htfa.local_weights_.shape)
-        assert htfa.local_weights_.shape[0] == n_tr * K,\
-            "Invalid result of HTFA! (wrong # element in local_weights)"
-        assert htfa.local_posterior_.shape[0] == htfa.prior_size,\
-            "Invalid result of HTFA! (wrong # element in local_posterior)"
+    for root in range(size):
```
mihaic: Is it not sufficient to run twice, once as master and once as worker?

xiazhu: the size here is 2 :) the command we use to test is mpirun -np 2 .....

mihaic: Why not make the code say `range(2)`?

xiazhu: i usually avoid hard-coded thing. And in this case, there is no difference to use "size".

yidawang: I think using `size` is proper in this case

mihaic: Could you please answer my original question? Is it not sufficient to run twice, once as master and once as worker? If it is sufficient, why have anything else but `2`?

xiazhu: It is not sufficient to run just twice, if "size" is larger than 2. We got report on every process, each process needs to be MPI root once.

mihaic: You are right. The MPICH `mpiexec` manual says:
`mpiexec returns the maximum of the exit status values of all of the processes created by mpiexec.`
http://www.mpich.org/static/docs/latest/www1/mpiexec.html

However, we can use the `-print-all-exitcodes` option to get all values:
```
mihai@alb:~$ mpiexec -n 3 -print-all-exitcodes hostname
alb
alb
alb
[mpiexec@alb] Exit codes: [alb] 0,0,0
```
If one of the processes returns 0, the test passes. This way, we can always run the minimum amount of tests. In this case, we only have 2 code paths, 1 for the master, 1 for the workers, so 2 passes are enough, irrespective of the number of processes.

BTW, I referenced `mpiexec`, not `mpirun`, because this is what the MPI standard specifies:
https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node228.htm
Yida is working on getting Intel MPI 2017 working on Metacortex so we can switch to `mpiexec`.

mihaic:
Retest this please.

mihaic:
Retest this please.

mihaic:
Retest this please.

mihaic:
We added a different implementation in #103.

PR #101, opened 2016-09-14
Call `git clean` for docs before building

mihaic:
The docs build process leaves behind files that can prevent other branch builds
from succeeding.


yidawang:
Good by me

PR #102, opened 2016-09-14
Deterministic SRM

TuKo:
Added the Deterministic Shared Response Model to Functional Alignment methods:

- Added tests
- Modified examples (python & notebook) to run with both Probabilistic and Deterministic SRM.

### DIFF
--- a/brainiak/funcalign/srm.py
+++ b/brainiak/funcalign/srm.py
@@ -11,9 +11,9 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-"""Probabilistic Shared Response Model (SRM)
+"""Shared Response Model (SRM)
 
-This implementation is based on the following publications:
+The implementations are based on the following publications:
 
 .. [Chen2015] "A Reduced-Dimension fMRI Shared Response Model",
    P.-H. Chen, J. Chen, Y. Yeshurun-Dishon, U. Hasson, J. Haxby, P. Ramadge
@@ -39,7 +39,7 @@ from sklearn.utils import assert_all_finite
 from sklearn.utils.validation import NotFittedError
 
 __all__ = [
-    "SRM",
+    "SRM", "DetSRM"
 ]
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


mihaic:
@cameronphchen Could you please review this PR?

brainiak/funcalign/srm.py
```diff
@@ -11,9 +11,9 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-"""Probabilistic Shared Response Model (SRM)
+"""Shared Response Model (SRM)
 
-This implementation is based on the following publications:
+The implementations are based on the following publications:
```
mihaic: Could you please change the second reference to our Arxiv paper?
https://arxiv.org/abs/1608.04647

cameronphchen:
@mihaic @TuKo  Sure, I'll review this.

brainiak/funcalign/srm.py
```diff
@@ -425,3 +426,241 @@ def _srm(self, data):
...
+    """
+
+    def __init__(self, n_iter=10, features=50, rand_seed=0, verbose=False):
+        self.n_iter = n_iter
+        self.features = features
+        self.rand_seed = rand_seed
+        self.verbose = verbose
+        return
+
+    def fit(self, X, y=None):
```
cameronphchen: if y is not used, why not remove it?

brainiak/funcalign/srm.py
```diff
@@ -425,3 +426,241 @@ def _srm(self, data):
...
+
+    s_ : array, shape=[features, samples]
+        The shared response.
+
+    .. note::
+       The number of voxels may be different between subjects. However, the
+       number of samples must be the same across subjects.
+
+       The Deterministic Shared Response Model is approximated using the
+       Block Coordinate Descent (BCD) algorithm proposed in [Chen2015]_.
```
cameronphchen: Can we remove the _ after [Chen2015]?

brainiak/funcalign/srm.py
```diff
@@ -425,3 +426,241 @@ def _srm(self, data):
...
+
+    def fit(self, X, y=None):
+        """Compute the Deterministic Shared Response Model
+
+        Parameters
+        ----------
+        X : list of 2D arrays, element i has shape=[voxels_i, samples]
+            Each element in the list contains the fMRI data of one subject.
+
+        y : not used
```
cameronphchen: same here, if y is not used, why not remove it?

brainiak/funcalign/srm.py
```diff
@@ -425,3 +426,241 @@ def _srm(self, data):
...
+
+        objective : float
+            The objective function value.
+        """
+        subjects = len(data)
+        objective = 0.0
+        for m in range(subjects):
+            objective += \
+                (np.linalg.norm(data[m] - w[m].dot(s), 'fro')**2)\
+                * (0.5 / data[m].shape[1])
```
cameronphchen: do we need this? since data[m] for all subjects should have the same number of TRs. and this is checked at the beginning of entering DetSRM.

brainiak/funcalign/srm.py
```diff
@@ -425,3 +426,241 @@ def _srm(self, data):
...
+    .. note::
+       The number of voxels may be different between subjects. However, the
+       number of samples must be the same across subjects.
+
+       The Deterministic Shared Response Model is approximated using the
+       Block Coordinate Descent (BCD) algorithm proposed in [Chen2015]_.
+
+       This is a single node version.
+
+       The run-time complexity is :math:`O(I (V T K))` and the memory
```
cameronphchen: Is this the complexity for matrix multiplication only? We would also need to incorporate the complexity of SVD.

tests/funcalign/test_srm.py
```diff
@@ -103,3 +114,104 @@ def test_can_instantiate():
...
+    # Check that it does not run with different number of samples (TRs)
+    S2 = S[:,:-2]
+    X.append(Q.dot(S2))
+    try:
+        model.fit(X)
+        assert True, "Success running DetSRM with different number of samples!"
+    except ValueError:
+        print("Catched Exception number 2: different number of samples per subject")
+
+
```
cameronphchen: Can we add a test with the synthetic data to make sure the output of DetSRM is consistent. To be specific, a test to check that the output Ws are all orthonormal and W_i*S are close to input X within a threshold delta. This doesn't have to be in this PR but it's necessary to have one in case some future editing break the code.

brainiak/funcalign/srm.py
TuKo: This is the interface for fit in BaseEstimator in scikit-learn.

brainiak/funcalign/srm.py
mihaic: That creates a link in the HTML version:
http://pythonhosted.org/brainiak/brainiak.funcalign.html

brainiak/funcalign/srm.py
TuKo: Will change it.

brainiak/funcalign/srm.py
TuKo: Correcting this one...

tests/funcalign/test_srm.py
TuKo: I've added the orthogonality check for both SRM and DetSRM. 
Added also the other test, but it is difficult to say what is the right delta.

TuKo:
@cameronphchen, please check the modifications and let us know if everything is fine.

PR #105, opened 2016-09-17
Use `setup.cfg` for `flake8` options

mihaic:
Ignore all violations ignored by `flake8` by default.

Use `--config` option to make sure `setup.cfg` is the only configuration
file used and other files users might have are ignored.

### DIFF
--- a/run-checks.sh
+++ b/run-checks.sh
@@ -17,5 +17,5 @@
 set -e
 set -o pipefail
 
-flake8 --ignore=W503 --max-complexity=10 brainiak
+flake8 --config setup.cfg brainiak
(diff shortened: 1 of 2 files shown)


run-checks.sh
```diff
@@ -17,5 +17,5 @@
 set -e
 set -o pipefail
 
-flake8 --ignore=W503 --max-complexity=10 brainiak
+flake8 --config setup.cfg brainiak
```
yidawang: why do we not need to ignore W503 any more?

mihaic: We do. It is ignored by default by flake8:
https://github.com/PyCQA/flake8/blob/master/src/flake8/defaults.py#L5

PR #116, opened 2016-09-27
Improvement to BRSA with example now

lcnature:
(1) added example
(2) fixed a bug in GP prior. Thanks @TuKo .
(3) added a weak Gaussian prior over log(SNR) when GP smoothness prior is not imposed. In fact, I still do not get the error that @TuKo  got with the random seed. But I think such a prior would prevent log(SNR) from going to crazy values.
(4) split 2 tests into 3 tests.
(5) other minor changes.


brainiak/reprsimil/brsa.py
```diff
@@ -248,7 +251,8 @@ def fit(self, X, design, scan_onsets=None, coords=None,
         assert_all_finite(design)
         assert design.ndim == 2,\
             'The design matrix should be 2 dimension ndarray'
-        assert np.all(np.std(design, axis=0) >= 0) or self.pad_DC,\
+        assert (not np.all(np.std(design, axis=0) > 0) and self.pad_DC)\
```
TuKo: Please, use exceptions and not asserts.
if bad condition is true:
    raise ValueError("message")  % or any other type of Error.

TuKo:
@lcnature please run nbstripout to your notebook and commit the file again.

brainiak/reprsimil/brsa.py
lcnature: Thanks @TuKo What is the advantage of using raise instead of assert?
It increases complexity. I used to change all the raise to assert for the complexity tests.

mihaic: Javier would prefer to use explicit exceptions for input checking instead of asserts. We discussed this face to face. There are advantages to doing that (too long discussion for PR comments). There are also advantages to using asserts. We can leave the code as it is.

TuKo: Agree.

lcnature:
@TuKo Can you be a bit more specific on "please use descriptive name of the example. don't say what algorithm are you showcasing, but what the example is doing."?
Do you mean removing the part of standard RSA for comparison? Or just change the name of the file?

TuKo:
Sorry, my mistake. use lowercase and underscores. if you can add a word or two for the description better,
something like this:
brsa_image_classification_example.ipynb

lcnature:
Thanks! Done

mihaic:
@lcnature Everything looks good. Please update your branch and I will merge the PR.

PR #125, opened 2016-10-02
minor bug fixed in logging debug output of fcma voxel selection

yidawang:
As the title. It is easy, only about a minor error in the logging output


TuKo:
Please, can you describe the bug or give an issue # (if exist)?
What was happening?

yidawang:
Sure. It is a typo in the displaying messages shown via logging debug, nothing related to the real computation. It is supposed to show the number of processed voxels stored in `task[1]`, however, I accidentally put it as `task[1] - task[0]`, where task[0] stores the starting id of the processed voxels.

PR #131, opened 2016-10-11
Add simulator

CameronTEllis:


setup.py
```diff
@@ -100,6 +100,8 @@ def build_extensions(self):
             ext.extra_link_args = opts
         build_ext.build_extensions(self)
 
+__pkgdata__ = {u'brainiak/utils/grey_matter_mask': [u'*.npy']}
```
mihaic: Please move this dictionary to the `setup` function; there is no need for the extra `__pkgdata__` variable.
Also, remove the `u` string prefixes. They are not necessary in Python 3, the only version we support.
Finally, the dictionary entry should be `'brainiak.utils': ['grey_matter_mask.npy']`.

PR #133, opened 2016-10-12
Add simulator

CameronTEllis:


examples/utils/simulator_example.py
```diff
@@ -0,0 +1,131 @@
...
+import logging
+
+import numpy as np
+from brainiak.utils import simulator as sim
+import matplotlib.pyplot as plt
+from mpl_toolkits.mplot3d import Axes3D as temp
+
+logger = logging.getLogger(__name__)
+
+temp.__module__  # Do in order to remove the pep8 error generated above
```
mihaic: There is a standard way to ignore Flake8 errors:
http://flake8.pycqa.org/en/latest/user/ignoring-errors.html
In this case, the import line should be:
```from mpl_toolkits.mplot3d import Axes3D  # noqa: F401```

PR #138, opened 2016-10-21
[sphinx] Set version using pkg_resources

mihaic:
Based on the latest recommendation from `setuptools_scm`, but using the
full version, as before:
https://github.com/pypa/setuptools_scm/issues/84

I removed the variable comments because they are no longer up to date.
They can be seen in the official documentation.


mihaic:
@xoltar Does this look good to you?

Issue #136, opened 2016-10-17
Add decision value/classifier confidence output to (real-time) FCMA


jhutchin:
Real-time FCMA output currently only includes binary classifier guess information. Additionally having access to the classifier confidence (e.g. decision value output in SVM) is critical.

yidawang:
On it

mihaic:
Fixed by #139.

PR #139, opened 2016-10-24
Fcma decision value

yidawang:
code refactoring of FCMA Classifier and addressing issue #136

### DIFF
--- a/brainiak/fcma/classifier.py
+++ b/brainiak/fcma/classifier.py
@@ -60,13 +60,13 @@ class Classifier(BaseEstimator):
 
     training_data_: 2D numpy array in shape [num_samples, num_features]
         default None
-        training_data_ is None except clf is SVM.SVC with precomputed kernel,
+        training_data\_ is None except clf is SVM.SVC with precomputed kernel,
         in which case training data is needed to compute
         the similarity vector for each sample to be classified
 
     test_data_: 2D numpy array in shape [num_samples, num_features]
         default None
-        test_data_ is set after a prediction is called,
+        test_data\_ is set after a prediction is called,
         so that the test data does not need to be regenerated in the
         subsequent operations, e.g. getting decision values of the prediction
 
@@ -97,7 +97,8 @@ class Classifier(BaseEstimator):
 
         Returns
         -------
-        corr_data: the correlation data in shape [len(X), num_voxels, num_voxels]
+        corr_data: the correlation data
+                    in shape [len(X), num_voxels, num_voxels]
         """
         num_samples = len(X)
         assert num_samples > 0, \
@@ -142,11 +143,13 @@ class Classifier(BaseEstimator):
 
         Parameters
         ----------
-        corr_data: the correlation data in shape [num_samples, num_voxels, num_voxels]
+        corr_data: the correlation data
+                    in shape [num_samples, num_voxels, num_voxels]
 
         Returns
         -------
-        corr_data: the normalized correlation data in shape [num_samples, num_voxels, num_voxels]
+        corr_data: the normalized correlation data
+                    in shape [num_samples, num_voxels, num_voxels]
         """
         # normalize if necessary
         if norm_unit > 1:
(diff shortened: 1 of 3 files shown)


brainiak/fcma/classifier.py
```diff
@@ -199,82 +304,54 @@ def predict(self, X):
...
+        if has_test_data:
+            assert len(X) == len(self.test_data_), \
+                'the number of samples of the input data X does not match ' \
+                'the number of samples of the test data'
+        else:
+            # generate the test_data first
+            # correlation computation
+            corr_data = self._prepare_corerelation_data(X)
+            # normalization
+            corr_data = self._normalize_correlation_data(corr_data, len(X))
```
mihaic: Name reuse.

brainiak/fcma/classifier.py
```diff
@@ -133,16 +133,121 @@ def fit(self, X, y):
...
+        assert len(X) == len(y), \
+            'the number of samples does not match the number labels'
+        num_samples = len(X)
+        num_voxels = X[0].shape[1]  # see assumption above
+        self.num_voxels_ = num_voxels
+        self.num_samples_ = num_samples
+        # correlation computation
+        corr_data = self._prepare_corerelation_data(X)
+        # normalization
+        corr_data = self._normalize_correlation_data(corr_data,
```
mihaic: You should avoid reusing names. How about something like `normalized_corr_data`?

brainiak/fcma/classifier.py
```diff
@@ -199,82 +304,54 @@ def predict(self, X):
...
+            on all test samples
+        has_test_data: Boolean value, default False
+            specify if the test_data has been generated
+
+        Returns
+        -------
+        confidence: the predictions confidence values of X, in shape [len(X),]
+        """
+        if has_test_data:
+            assert len(X) == len(self.test_data_), \
```
mihaic: This is not necessarily a sufficient test. Why not store `X` as `self.raw_test_data_` and compare to that?

brainiak/fcma/classifier.py
```diff
@@ -199,82 +304,54 @@ def predict(self, X):
...
+            if len(X) > 1: normalization is done
+            on all test samples
+        has_test_data: Boolean value, default False
+            specify if the test_data has been generated
+
+        Returns
+        -------
+        confidence: the predictions confidence values of X, in shape [len(X),]
+        """
+        if has_test_data:
```
yidawang: Even if the classifier has `test_data_`, it may be from the previous input data, not the current input data from which we'd like to compute decision values. So it requires the user to decide if the input data has been transformed and predicted.

brainiak/fcma/classifier.py
yidawang: `X` is the input activity data, `self.test_data_` is the correlation data, here we only compare if these two contains the same number of samples as a sanity check.

mihaic: I understand what `X` is, that is why I proposed calling storing it as `self.raw_test_data_`.

In the current proposal, if the user calls `decision_function(X1, False)` and then mistakenly calls `decision_function(X2, True)` (instead of `decision_function(X2, False)`), the result will be wrong if `len(X1) == len(X2)`.

With my proposal, the interface will be simpler (no more `has_test_data` parameter) and the check will ensure old test data is not reused.

yidawang: I get it. The drawback is that `X` can be large, comparing its content takes time.

brainiak/fcma/classifier.py

brainiak/fcma/classifier.py

brainiak/fcma/classifier.py
```diff
@@ -133,43 +139,154 @@ def fit(self, X, y):
...
             corr_data = corr_data.reshape(1,
                                           num_samples,
                                           num_voxels * num_voxels)
-            fcma_extension.normalization(corr_data, self.epochs_per_subj)
+            fcma_extension.normalization(corr_data, norm_unit)
             corr_data = corr_data.reshape(num_samples, num_voxels, num_voxels)
             logger.debug(
                 'normalization done'
             )
+        return corr_data
```
mihaic: Please do not use `corr_data` for different types of data.

examples/fcma/classification.py
```diff
@@ -48,4 +48,5 @@
     #joblib.dump(clf, 'model/logistic.pkl')
     #clf = joblib.load('model/svm.pkl')
     print(clf.predict(test_data))
+    print(clf.decision_function(test_data, has_test_data=True))
```
mihaic: No need for `has_test_data` anymore.

yidawang: done

brainiak/fcma/classifier.py

PR #140, opened 2016-10-26
minor bug fixed

yidawang:
a minor caused by renaming `normalized_corr_data` in `_normalize_correlation_data` method. It should return the original `corr_data` if no normalization is performed.


brainiak/fcma/classifier.py
```diff
@@ -171,6 +171,8 @@ def _normalize_correlation_data(self, corr_data, norm_unit):
             logger.debug(
                 'normalization done'
             )
+        else:
```
mihaic: Could you please document `norm_unit` in the docstring?

PR #141, opened 2016-10-26
[eventseg] Fix binary module name issue

mihaic:
Python started complaining about `brainiak.eventseg.utils`:
'dynamic module does not define module export function (PyInit_utils)'
https://travis-ci.org/IntelPNI/brainiak/jobs/170847241

Renaming it `brainiak.eventseg._utils`, in accordance to the Cython source file
name, fixes the issue.


mihaic:
@yidawang Please have a look.

Issue #134, opened 2016-10-12
Documentation issues with some macros (notes, math, etc)


TuKo:
There are several issues with the macros and the way documentation is being written.
Some math (latex) is rendered incorrectly: see [SSSRM class description](https://pythonhosted.org/brainiak/brainiak.funcalign.html#brainiak.funcalign.sssrm.SSSRM) for an example.

Also the [notes](https://pythonhosted.org/brainiak/brainiak.funcalign.html#brainiak.funcalign.srm.DetSRM) are not shown correctly 

The documentation is available [here](https://pythonhosted.org/brainiak/).

lcnature:
@TuKo It happens to my equations too. I think I just copied my latex text from my paper. How to check the rendering locally before pulling request?

TuKo:
@lcnature, you can create the documentation by calling `make` in the docs subfolder.
Then you open the file ./docs/_build/html/index.html with a browser.

mihaic:
I am working on a PR for this issue.

mihaic:
Fixed by PR #143.

PR #143, opened 2016-10-27
Fix math and notes

mihaic:
The most important problem was backslashes that were not escaped.

In the BRSA code, I also changed eta to epsilon to match the paper.

I also replaced a couple of duplicate equations with references.

For notes, the problem was that the `note` directive was mistaken for
another attribute or parameter. It needs to be separated by at least two
blank lines to be recognized as a directive. Instead I replaced it with
a Note section, which looks better in the raw docstring and looks the
same in HTML.

Fixes issue #134.

### DIFF
--- a/brainiak/funcalign/srm.py
+++ b/brainiak/funcalign/srm.py
@@ -70,10 +70,13 @@ def _init_w_transforms(data, features):
         A list with the number of voxels per subject.
 
 
-    .. note:: This function assumes that the numpy random number generator was
-       initialized.
+    Note
+    ----
 
-       Not thread safe.
+        This function assumes that the numpy random number generator was
+        initialized.
+
+        Not thread safe.
     """
     w = []
     subjects = len(data)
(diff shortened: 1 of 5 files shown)


mihaic:
@TuKo Happy?

brainiak/funcalign/srm.py
```diff
@@ -428,7 +433,7 @@ class DetSRM(BaseEstimator, TransformerMixin):
     Given multi-subject data, factorize it as a shared response S among all
     subjects and an orthogonal transform W per subject:
 
-    .. math:: X_i \\approx W_i S ,~for~all~i=1\dots N
+    .. math:: X_i \\approx W_i S, \\exists i=1 \\dots N
```
TuKo: Change \\exists by \\forall in SRM

brainiak/funcalign/srm.py
```diff
@@ -95,7 +98,7 @@ class SRM(BaseEstimator, TransformerMixin):
...
```
TuKo: Change \\exists by \\forall in SRM

mihaic:
Good catch, @TuKo. Done.

PR #147, opened 2016-11-08
[dev] Declare minimum Python version in setup.py

mihaic:
In v24.2.0, setuptools introduced support for declaring the minimum
Python version that the package is guaranteed to work with by means
of the `python_requires` keyword:
http://setuptools.readthedocs.io/en/latest/history.html#v24-2-0


mihaic:
@xoltar Does this make sense to you?

PR #157, opened 2017-01-12
add mvpa training and classification examples

yidawang:
Thanks for reviewing

### DIFF
--- a/examples/fcma/mvpa_classification.py
+++ b/examples/fcma/mvpa_classification.py
@@ -1,0 +1,139 @@
+#  Copyright 2016 Intel Corporation
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from sklearn import svm
+#from sklearn.linear_model import LogisticRegression
+import sys
+import os
+import logging
+from file_io import generate_epochs_info
+import numpy as np
+from scipy.stats.mstats import zscore
+import nibabel as nib
+#from sklearn.externals import joblib
+
+format = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
+# if want to output log to a file instead of outputting log to the console,
+# replace "stream=sys.stdout" with "filename='fcma.log'"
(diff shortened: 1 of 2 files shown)


examples/fcma/mvpa_classification.py
```diff
@@ -0,0 +1,139 @@
...
+#from sklearn.externals import joblib
+
+format = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
+# if want to output log to a file instead of outputting log to the console,
+# replace "stream=sys.stdout" with "filename='fcma.log'"
+logging.basicConfig(level=logging.INFO, format=format, stream=sys.stdout)
+logger = logging.getLogger(__name__)
+
+
+def prepare_data(raw_data, epoch_info):
```
mihaic: Isn't it better to make `MVPAVoxelSelector._preprocess_data` work with 2D data, make it public, and use it here instead of copying the code?

examples/fcma/mvpa_classification.py
```diff
@@ -0,0 +1,139 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from sklearn import svm
+#from sklearn.linear_model import LogisticRegression
+import sys
+import os
+import logging
+from file_io import generate_epochs_info
```
mihaic: You should consider moving `file_io` from examples to `brainiak.fcma`.

examples/fcma/mvpa_classification.py
```diff
@@ -0,0 +1,139 @@
...
+             and f.endswith(extension)]
+    for f in files:
+        img = nib.load(os.path.join(data_dir, f))
+        data = img.get_data()
+        # apply mask
+        data = data[mask, :]
+        raw_data.append(data)
+        logger.info(
+            'file %s is loaded, with data shape %s' % (f, data.shape)
+        )
```
mihaic: This code looks a lot like `file_io.read_activity_data`. Could you unify them?

examples/fcma/mvpa_classification.py
yidawang: makes sense. Let me do this later

examples/fcma/mvpa_classification.py
yidawang: `MVPAVoxelSelector._preprocess_data` has to be in 4D so as to use `Searchlight`

examples/fcma/mvpa_classification.py
yidawang: Yes, I am also thinking of it, too. Currently, there is a slight difference between these two: one producing `nTRs*nVoxels` matrix in row-major; the other producing `nVoxels*nTRs` matrix in row-major. I will simplify the code (including moving `file_io.py` into brainiak) after some more thinking.

mihaic:
As discussed, I will create an issue about the review comments to be addressed in a later PR by Yida.

PR #159, opened 2017-01-13
Corr comp

yidawang:
Add a high-performance correlation computation function using BLAS wrapped in Cython to `brainiak/fcma/util`. Running at least one order of magnitude faster than `np.corrcoef`.
```
Ours: 100 loops, best of 3: 14.2 ms per loop
np.corrcoef: 10 loops, best of 3: 153 ms per loop
```

### DIFF
--- a/brainiak/fcma/util.py
+++ b/brainiak/fcma/util.py
@@ -42,11 +42,12 @@ def compute_correlation(component1, component2):
     component1 = normalize_for_correlation(component1, 1)
     component2 = normalize_for_correlation(component2, 1)
     corr_data = np.empty((r1, r2), dtype=np.float32, order='C')
+    # blas routine is column-major
     blas.compute_single_matrix_multiplication('T', 'N',
                                               r1, r2, d1,
                                               1.0,
-                                              component1, d1,
                                               component2, d2,
+                                              component1, d1,
                                               0.0,
                                               corr_data,
                                               r2)
--- a/brainiak/fcma/util.py
+++ b/brainiak/fcma/util.py
@@ -1,0 +1,53 @@
+#  Copyright 2016 Intel Corporation
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


brainiak/fcma/util.py
```diff
@@ -0,0 +1,95 @@
...
+    return data
+
+
+def compute_correlation(set1, set2):
+    """compute correlation between two sets
+
+    Correlate the rows of set1 with the rows of set2.
+    If set1 == set2, it is auto-correlation computation
+    resulting in a symmetric correlation matrix.
+    The number of columns MUST agree between set1 and set2
```
mihaic: Could you please add a reference for the method of correlation computation you are using? Some Latex code could help too.

brainiak/fcma/util.py
```diff
@@ -0,0 +1,95 @@
...
+    shape = data.shape
+    data = zscore(data, axis=axis, ddof=0)
+    # if zscore fails (standard deviation is zero),
+    # set all values to be zero
+    data = np.nan_to_num(data)
+    data = data / math.sqrt(shape[axis])
+    return data
+
+
+def compute_correlation(set1, set2):
```
mihaic: Shouldn't the parameters be called `array1/2` or `matrix1/2`?

yidawang: they are essentially two matrices, depicting two sets of variables. I can use `matrix1/2`

brainiak/fcma/util.py
yidawang: done

brainiak/fcma/util.py
```diff
@@ -70,6 +70,19 @@ def compute_correlation(matrix1, matrix2):
 
     is the standard deviation of variable X
 
+    Reducing the correlation computation to matrix multiplication
+    and use BLAS GEMM API wrapped by Scipy can speedup the numpy build-in
```
mihaic: use->using, build-in->built-in

yidawang: done

brainiak/fcma/util.py
```diff
@@ -0,0 +1,118 @@
...
+# Authors: Yida Wang
+# (Intel Labs), 2017
+
+import numpy as np
+from . import cython_blas as blas
+from scipy.stats.mstats import zscore
+import math
+
+
+def normalize_for_correlation(data, axis):
```
mihaic: Is it necessary for this function to be public?

yidawang: May or may not be used by others in the future. But for now, let me make it private

PR #167, opened 2017-01-24
Fmrisim to generate data for fcma

CameronTEllis:


examples/fcma/generate_fcma_data.py
```diff
@@ -0,0 +1,164 @@
...
+# Generate the mask of the signal
+mask = sim.mask_brain(signal)
+
+# Mask the signal to the shape of a brain (attenuates signal according
+# to grey matter likelihood)
+signal *= mask
+
+# Iterate through the participants and store participants
+epochs = []
+for participantcounter in list(range(1, participants + 1)):
```
mihaic: No need for `list` here.

examples/fcma/generate_fcma_data.py
```diff
@@ -0,0 +1,164 @@
...
+logger = logging.getLogger(__name__)
+
+
+# Set the experimental parameters
+participants = 5
+epochs = 5
+dimensions = np.array([10, 10, 10])  # What is the size of the brain
+
+# Where will the data be stored?
+directory = 'examples/fcma/simulated/'
```
mihaic: This will not work as you expect if you execute the example from anywhere else but the root of the Git repository. You should test if `examples/fcma` exists. If not, exit and ask the user to pass the path as an argument; you can get it as `sys.argv[1]`.

brainiak/utils/fmrisim.py
```diff
@@ -574,40 +576,47 @@ def double_gamma_hrf(stimfunction,
...
     hrf_length = 30  # How long is the HRF being created
 
-    hrf = [0] * hrf_length  # How many seconds of the HRF will you model?
+    # How many seconds of the HRF will you model?
+    hrf = [0] * int(hrf_length / temporal_resolution)
 
-    for hrf_counter in list(range(0, hrf_length - 1)):
+    for hrf_counter in list(range(len(hrf) - 1)):
         # When is the peak of the two aspects of the HRF
         response_peak = response_delay * response_dispersion
```
lcnature: Hi Cameron, I think that response_delay, response_dispension and the two other terms on the right hand side of the line below do not change within the for loop. So the result of these two lines will be exactly the same in each iteration of the for loop. Therefore, it seems that they can be taken to before the for loop for slightly faster speed.

brainiak/utils/fmrisim.py
```diff
@@ -574,40 +576,47 @@ def double_gamma_hrf(stimfunction,
...
 
+    # Decimate the signal function so that it only has one element per TR
+    signal_function = signal_function[0::int(tr_duration * 1000)]
+
     # Cut off the HRF
-    signal_function = signal_function[0:len(stimfunction)]
+    signal_function = signal_function[0:int(len(stimfunction) / 1000)]
 
     # Scale the function so that the peak response is 1
     signal_function = signal_function / np.max(signal_function)
```
lcnature: Hi @CameronTEllis ,
Is it possible to make this scaling optional? In my usage I don't want to scale the signal. I could of course write my own function. But if this is optional, I can call your function.
Thanks!

PR #169, opened 2017-01-27
Matnorm models base

mshvartsman:
Code for matrix variate normal models for fMRI analysis. Currently includes regression and BRSA with the following covariances: constant, identity, isotropic, diagonal, full rank unconstrained, and a shim to use all the kernels in `GPflow`. All of these covariances can be used for space covariance, time covariance, or even as the structure assumption for the RSA matrix.


mshvartsman:
Some work to do before it's ready to merge, but it's mostly linter cleanup and other similar bookkeeping so I thought we could kick this off for code review purposes. Paging @lcnature and @cameronphchen for the algo part and anyone else for the software side.

mihaic:
If some of the linter codes are not clear, let me know.

mshvartsman:
Linter codes are clear. I've avoided fixing some line length stuff where I don't think it would help (e.g. going from a 82-char line to a weirdly broken up set of lines). 

Also: can someone advise on how to deal with the `define_scope` decorators in `helpers.py`? I got it from a guy's blog post, not clear what the license there is.

mihaic:
I also asked the author about including the code here. Lets hope he agrees.

TuKo:
I will look at the code in the incoming days.

brainiak/matnorm_models/base.py
```diff
@@ -0,0 +1,229 @@
...
+        logdet = -Q.logdet + sigma.logdet +\
+            2 * tf.reduce_sum(tf.log(tf.matrix_diag_part(i_qf_cholesky)))
+
+        # A' Sigma^{-1}
+        Atrp_Sinv = tf.matmul(A, sigma.Sigma_inv, transpose_a=True)
+        # (Q - A' Sigma^{-1} A)^{-1} A' Sigma^{-1}
+        prod_term = tf.cholesky_solve(i_qf_cholesky, Atrp_Sinv)
+
+        solve = tf.matmul(sigma.Sigma_inv +
+                          sigma.Sigma_inv_x(tf.matmul(A, prod_term)), x)
```
TuKo: If x is a vector (or a small column size matrix, then you should consider multiplying by x first.
(Sinv + Sinv*(A*Prod))*X  = Sinv*(I + A*Prod)*x.
So you can call once to Sigma_inv_x()
Also you can consider the cholesky solve above  for x+A*Prod_x and you can compute first Prod_x =chol_solve(i_qf_chol, A'*sigma_inv_x(x))

Everything depends on the size of x and the matrices. 
A better way to do this is using [a conditional structure](https://www.tensorflow.org/versions/master/api_docs/python/control_flow_ops/control_flow_operations#cond).

brainiak/matnorm_models/noise_covs.py
```diff
@@ -0,0 +1,268 @@
...
+        pass
+
+    @abc.abstractmethod
+    def Sigma_inv_x(self, X):
+        """Given this Sigma and some X, compute :math:`Sigma^{-1} * x`
+        """
+        pass
+
+    @define_scope
+    def Sigma_inv(self):
```
TuKo: What is the reason to define Sigma_inv? (you should always use sigma_inv_x() and enforce it by not having sigma_inv)

brainiak/matnorm_models/noise_covs.py
```diff
@@ -0,0 +1,268 @@
+import tensorflow as tf
+import numpy as np
+import abc
+from .helpers import define_scope, xx_t
+
+
+class NoiseCovBase:
```
TuKo: I think that the cholesky factor should always be a property of these covariances. For your consideration...

brainiak/matnorm_models/noise_cov_gp.py
```diff
@@ -0,0 +1,53 @@
+from .noise_covs import NoiseCovBase
```
TuKo: This code should be inside the file noise_covs.py

brainiak/matnorm_models/noise_cov_kron.py
```diff
@@ -0,0 +1,124 @@
+import tensorflow as tf
```
TuKo: This code should be inside the file noise_covs.py

brainiak/matnorm_models/noise_covs.py
```diff
@@ -0,0 +1,268 @@
+import tensorflow as tf
+import numpy as np
+import abc
+from .helpers import define_scope, xx_t
+
```
TuKo: Why do you call NoiseCov* to these matrices?
What is the Noise part in them? 
It seems more realistic to call them Covariance* or Cov*. (same for the filename).

Is there any reason to initialize most of them randomly? why not to offer an initialization value option and if it is None then use the random values?

brainiak/matnorm_models/base.py
```diff
@@ -0,0 +1,229 @@
...
+            self.sess.run(optfun, feed_dict=feed_dict)
+            past_loss = current_loss
+            current_loss = self.sess.run(loss, feed_dict=feed_dict)
+            max_abs_current_grad = self.sess.run(max_abs_current_grad_op,
+                                                 feed_dict=feed_dict)
+
+            # check tolerances
+            if abs((current_loss - past_loss) /
+                    max(current_loss, past_loss)) < loss_tol:
+                print('loss tolerance reached on iter %i, %f, stopping' %
```
mihaic: Indeed, we do not use `print` in BrainIAK. This used to be checked automatically, but the tool broke:
https://github.com/JBKahn/flake8-print/pull/17

brainiak/matnorm_models/helpers.py
```diff
@@ -0,0 +1,73 @@
+import functools  # https://danijar.com/structuring-your-tensorflow-models/
+import tensorflow as tf
+from numpy.linalg import cholesky
+from scipy.stats import norm
+
```
mihaic: It could go to "brainiak/utils/tensorflow.py", except for the `rmn`, which could stay in this file, renamed "utils.py".

In any case, we cannot merge this code until we hear from @danijar.

brainiak/matnorm_models/base.py
```diff
@@ -0,0 +1,229 @@
...
+
+        # (Qinv - A' Sigma^{-1} A)
+        i_qf_cholesky = tf.cholesky(Q.Sigma - tf.matmul(A,
+                                    sigma.Sigma_inv_x(A), transpose_a=True))
+
+        logdet = -Q.logdet + sigma.logdet +\
+            2 * tf.reduce_sum(tf.log(tf.matrix_diag_part(i_qf_cholesky)))
+
+        # A' Sigma^{-1}
+        Atrp_Sinv = tf.matmul(A, sigma.Sigma_inv, transpose_a=True)
```
mshvartsman: I might be misunderstanding your suggestion, but: `sigma.Sigma_inv` is defined a `self.Sigma_inv_x(tf.diag(tf.ones([self.size])))` in the covariance base, but can be overriden with something more efficient for a particular covariance. If I just write `self.Sigma_inv_x(tf.diag(tf.ones([self.size])))` here then we lose that option. What is the reason to change?

brainiak/matnorm_models/base.py
mshvartsman: In general X can be quite large (it is either TRs by voxels or sometimes voxels by TRs). Saving one solve by factoring out Sinv makes sense. The other suggestion (`Prod_x =chol_solve(i_qf_chol, A'*sigma_inv_x(x))`) is dependent on the size of x as you say. By my quick check (below, might be off) if x has many more columns than rows then what I wrote is best, as n approaches m then your suggestion works better. For what we're using this for so far (marginalizing over the TR dimension), voxels >> TRs and therefore the current option should be better, but agreed that conditional structure based on sizes is ideal. Can we register this as an issue for work after the PR? 


```
# X is m*n, Q is k*k, A is m*k, Sigma is m*m

Option 1: 
# A' Sigma^{-1}: km^2 plus the cost of Sigma_inv
Atrp_Sinv = tf.matmul(A, sigma.Sigma_inv, transpose_a=True)
# (Q - A' Sigma^{-1} A)^{-1} A' Sigma^{-1} # k^2 for the solve 
prod_term = tf.cholesky_solve(i_qf_cholesky, Atrp_Sinv)
# m^2k for the inner matmul, m for the addition, m^2 for the solve, m^2n for the outer matmul
solve = tf.matmul(sigma.Sigma_inv_x(scaled_I(1.0, m)+
                          tf.matmul(A, prod_term)), x)

# total: 2km^2 + k^2 + m^2 + nm^2 + m

Option 2:
# m^2 for the solve, kmn for the mult, k^2 for other solve
# (Q - A' Sigma^{-1} A)^{-1} A' Sigma^{-1} X 
prod_term = tf.cholesky_solve(i_qf_cholesky, tf.matmul(A, sigma.Sigma_inv(x), transpose_a = True)
# mkn for inner matmul, m for addition, m^2 for solve
solve = sigma.Sigma_inv_x(scaled_I(1.0, m) + tf.matmul(A, prod_term))

# total: 2kmn + 2m^2 + k^2 + m
```

brainiak/matnorm_models/base.py
```diff
@@ -0,0 +1,229 @@
...
+        """Construct logp from the solves and determinants.
+        """
+        log2pi = 1.8378770664093453
+
+        denominator = - rowsize * colsize * log2pi -\
+            colsize * logdet_row - rowsize * logdet_col
+        numerator = - tf.trace(tf.matmul(solve_col, solve_row))
+        return tf.reduce_sum(0.5 * (numerator + denominator))
+
+    def matnorm_logp(self, x, row_cov, col_cov):
```
mshvartsman: In general I would agree. However, the way MatnormModelBase is subclassed makes this less clear from context. For example, in matnom-BRSA we currently call `self.matnorm_logp_marginal_row`. If this instead were `self.logp_marginal_row` it is less clear what it is the log likelihood for exactly (it is not for BRSA, it is specifically for a matrix varaite normal with a marginalization term in the row covariance). 

On reflection I wonder if what this is telling me is that I those likelihoods should live in their own namespace and not be class methods (so instead of `self.matnorm_logp` we would `from matnorm_likelihoods import matnorm_logp` and then just use it. Thoughts?

brainiak/matnorm_models/base.py
```diff
@@ -0,0 +1,229 @@
...
+        solve_row = row_cov.Sigma_inv_x(x)
+        logdet_row = row_cov.logdet
+
+        solve_col, logdet_col = self.solve_det_marginal(tf.transpose(x),
+                                                        col_cov, marg, marg_cov)
+
+        return self._mnorm_logp_internal(colsize, rowsize, logdet_row,
+                                         logdet_col, solve_row, solve_col)
+
+    def matnorm_logp_conditional_row(self, x, row_cov, col_cov, cond, cond_cov):
```
mshvartsman: If they stay as methods in MatnormBase, agreed. But see above about possibly moving them out, in which case I think they're fine returning TF objects (possibly even following tf.contrib.distributions API).

brainiak/matnorm_models/regression.py
```diff
@@ -0,0 +1,167 @@
...
+    space_noise_cov : subclass of NoiseCovBase
+        Voxel noise covariance class following NoiseCovBase interface.
+    learnRate : real, default=0.01
+        Step size for the Adam optimizer
+
+    """
+    def __init__(self, n_c, n_v, time_noise_cov, space_noise_cov,
+                 learnRate=0.01):
+
+        self.n_c, self.n_v, self.time_noise_cov, self.space_noise_cov =\
```
mshvartsman: I've intentionally avoided having a base class `__init__` to be maximally agnostic about derived models -- right now the base class is basically a collection of methods without any state. For example, if we were to implement SRM in this framework we wouldn't have time and space cov, we'd have subject-space and time covs, and we wouldn't have n_c. If we were to replace the optimizer with the "flip-flop" coordinate ascent algo for matrix normal covariance estimation, we wouldn't have the same optimizer node -- and so on and so forth. I do want to make more commitments to the base API, but want to do them thoughtfully based on more than just these two simple models. Maybe I'm overthinking it at this early stage, though?

brainiak/matnorm_models/regression.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        return self.optimizer.minimize(-self.train_logp,
+                                       var_list=self.train_variables)
+
+    @define_scope
+    def train_logp(self):
+        """ Log likelihood of model (internal)
+        """
+        y_hat = tf.matmul(self.X, self.beta)
+        resid = self.Y - y_hat
```
mshvartsman: `y_hat` is the estimate, yes, and `resid` is residual. I can rename both (though hat for estimate is pretty standard in stats I think).

brainiak/matnorm_models/regression.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+
+        y_predicted = tf.matmul(self.X, self.beta)
+
+        self.y_predicted_ = y_predicted.eval(session=self.sess, 
+                                             feed_dict={self.X: X})
+
+        return self.y_predicted_
+
+    def calibrate(self, Y):
```
mshvartsman: https://en.wikipedia.org/wiki/Calibration_(statistics)#In_regression is a term for finding X from Y after having fit a regression model Y ~ XB. Aka "inverse regression" or what fMRI people would call a decoder, but I didn't want to call it `decode` because this is also what fMRI people would call an encoding model. I'm open to other suggestions.

brainiak/matnorm_models/noise_covs.py
mshvartsman: I think mentioned above, but I'm doing this because sometimes Sigma_inv is trivial. Is it always the case that if Sigma_inv is trivial then Sigma_inv_x(I) is also trivial? Or are you saying that we should check inside Sigma_inv_x for X being identity?

brainiak/matnorm_models/noise_cov_gp.py
mshvartsman: I've separated it intentionally since it's an optional shim for using GPflow kernels and not another covariance. Can you provide more justification for why we should combine them? (I do agree about combining the kronecker covariance with the rest).

mshvartsman:
Thank you for the detailed comments! Assume that if I haven't responded to a specific item above, I am in agreement and working on the changes requested.

brainiak/matnorm_models/base.py
TuKo: You are doing A * Sigma_inv,  it should be better to use Sigma.Sigma_inv_x(A.T) in line 95.

brainiak/matnorm_models/noise_cov_kron.py
```diff
@@ -0,0 +1,124 @@
...
+import numpy as np
+from .helpers import define_scope
+from .noise_covs import NoiseCovBase
+
+
+def tf_solve_lower_triangular_kron(L, y):
```
narayanan2004: We are using tf.matrix_triangular_solve() for the individual matrices. Maybe I am not getting what you mean here.

brainiak/matnorm_models/noise_cov_kron.py
```diff
@@ -0,0 +1,124 @@
...
+    '''
+    Calculate L x = y where L = kron(L[0], L[1] .. L[n-1])
```
narayanan2004: I will use "Solve" instead of "Calculate"

brainiak/matnorm_models/noise_cov_kron.py
TuKo: Sorry, now I see that you are passing L[1:] in the recursive call.

mjanderson09:
I notice that in brsa.py __init__ function there is a parameter "structured_RSA_cov " which isn't included in the documentation in the lines above the function.

mshvartsman:
I'm going to close this for now while we redesign matnormal. Will incorporate comments given here when I reopen.

PR #171, opened 2017-01-27
fcma cross validation using model selection

yidawang:
add score method to Classifier so that `sklearn.model_selection` can be used
add cross validation examples using model selection
update the test code


brainiak/fcma/classifier.py
```diff
@@ -252,14 +252,40 @@ def _prepare_test_data(self, corr_data):
...
+        the resulting correlation is large, the kernel matrix will be computed
+        portion by portion to save memory usage (the portion size is specified
+        in self.num_processed_voxels.
+
+        Parameters
+        ----------
+        X: a list of numpy array in shape [num_TRs, num_voxels]
+            X contains the activity data filtered by top voxels
+            and prepared for correlation computation.
+            assuming all elements of X has the same num_voxels value
```
mihaic: Typo: "has". Also, please write sentences, with capitalization and punctuation.

brainiak/fcma/classifier.py
```diff
@@ -502,3 +528,35 @@ def decision_function(self, X=None):
...
+
+        Parameters
+        ----------
+        X: a list of numpy array in shape [num_TRs, self.num_voxels\_]
+            Test samples.
+            It is not used in sklearn.svm.SVC with precomputed
+            kernel when the kernel matrix is computed portion by portion.
+        y : labels, len(X) equals len(y)
+        sample_weight : 1D array in shape [n_samples], optional
+            Sample weights.
```
mihaic: Does `num_TRs == n_samples`? If so, why not use the same name?

brainiak/fcma/classifier.py
```diff
@@ -502,3 +528,35 @@ def decision_function(self, X=None):
...
+
+        NOTE: This function will not work in sklearn.svm.SVC with precomputed
+        kernel when the kernel matrix is computed portion by portion.
+
+        Parameters
+        ----------
+        X: a list of numpy array in shape [num_TRs, self.num_voxels\_]
+            Test samples.
+            It is not used in sklearn.svm.SVC with precomputed
+            kernel when the kernel matrix is computed portion by portion.
```
mihaic: You say at the beginning of the docstring that the function does not work in this case. Here you say `X` is not used, implying the function does work. Can you please clarify?

brainiak/fcma/classifier.py
```diff
@@ -502,3 +528,35 @@ def decision_function(self, X=None):
...
+
+    def score(self, X, y, sample_weight=None):
+        """ returns the mean accuracy on the given test data and labels
+
+        NOTE: This function will not work in sklearn.svm.SVC with precomputed
+        kernel when the kernel matrix is computed portion by portion.
+
+        Parameters
+        ----------
+        X: a list of numpy array in shape [num_TRs, self.num_voxels\_]
```
mihaic: You refer to an actual variable, `self.num_voxels_`. However, you use `num_TRs` as a placeholder next to it, simply to indicate that TRs should be the first axis. I find this a bit confusing. How about choosing names that are clearly not variable names, e.g., `[TRs, voxels]`, and explaining that voxels must be `self.num_voxels_`?

Clearly this is a larger discussion, affecting all our documentation:
https://mail.scipy.org/pipermail/scipy-dev/2013-January/018261.html
There is no conclusion.

I brought it up here because I noticed the difference in roles between the two shape element names.

brainiak/fcma/classifier.py
```diff
@@ -252,14 +252,40 @@ def _prepare_test_data(self, corr_data):
...
         return data
 
-    def _compute_similarity_matrix_in_portion(self, X):
+    def _compute_kernel_matrix_in_portion(self, X):
+        """ compute kernel matrix for sklearn.svm.SVC with precomputed kernel
+
+        The method generates the kernel matrix (similarity matrix) for
+        sklearn.svm.SVC with precomputed kernel. It first computes
+        the correlation from X, then normalizes the correlation if needed,
+        and finally compute the kernel matrix. It is worth noting that if
```

brainiak/fcma/classifier.py

brainiak/fcma/classifier.py
yidawang: Fair enough. Changed.

brainiak/fcma/classifier.py
yidawang: Oh, this is fixed in my second commit. I fixed the inconsistency.

brainiak/fcma/classifier.py
yidawang: no, they are different. `n_sample` is the number of epochs.

PR #172, opened 2017-01-31
Fcma scalable searchlight

yidawang:
In the activity-based voxel selection using Searchlight, preprocess the data subject by subject while reading in, instead of preprocessing in batch after reading all the data, so as to save memory to handle high-resolution datasets (e.g. 91x109x91 brain volume).

Thanks in advance for reviewing!


brainiak/fcma/mvpa_voxelselector.py
```diff
@@ -38,96 +37,37 @@ class MVPAVoxelSelector:
     Parameters
     ----------
 
-    raw\_data: list of 4D array
-        each element of the list is the raw data of a subject,
-        in the shape of [x, y, z, t]
+    data: 4D array in shape [brain 3D + epoch]
+        contains the averaged and normalized brain data epoch by epoch.
+        It is generated by io.prepare_searchlight_mvpa_data
```
mihaic: This would be clearer if you specified the path as in an import statement, i.e., `.io.prepare_searchlight_mvpa_data`.

brainiak/fcma/mvpa_voxelselector.py
```diff
@@ -38,96 +37,37 @@ class MVPAVoxelSelector:
...
 
-    epoch\_info: list of tuple (label, sid, start, end).
-        label is the condition labels of the epochs;
-        sid is the subject id, corresponding to the index of raw_data;
-        start is the start TR of an epoch (inclusive);
-        end is the end TR of an epoch(exclusive).
-        Assuming len(labels) labels equals the number of epochs and
-        the epochs of the same sid are adjacent in epoch_info
+    labels: 1D array
+        contains the labels of the epochs, extracted from epoch\_info
```
mihaic: `epoch_info` is no longer defined in this module. Make a reference to its definition in `.io` or, better yet, remove it, since it is superfluous.

brainiak/fcma/mvpa_voxelselector.py
yidawang: done

brainiak/fcma/mvpa_voxelselector.py
yidawang: You are right, I removed most of the corresponding info but left this one. Fixed.

Issue #149, opened 2016-11-23
Refactor code to remove Pathos


mihaic:
The Multiprocess package, used by Pathos, should work:
https://pypi.python.org/pypi/multiprocess

Mike tried, but had trouble limiting the number of processes created.

Even better, look into using standard library multiprocessing directly with the Dill serializer (also used by Pathos):
https://pypi.python.org/pypi/dill

Mike tried, but could not obtain the same performance because the serialization/deserialization code was called unnecessarily often.

- [x] Searchlight
- [x] FCMA

mihaic:
@mjanderson09, can you please also try the pool context manager in the searchlight code, like PR #177?

mihaic:
Fixed by #225.

PR #176, opened 2017-02-03
Fcma parallelize cv

yidawang:
Use `pathos.multiprocessing` to parallelize the cross-validation step of correlation-based voxel selection when `sklearn.svm.SVC` with precomputed kernel is used as the classifier.

Thanks in advance for reviewing!

### DIFF
--- a/brainiak/fcma/voxelselector.py
+++ b/brainiak/fcma/voxelselector.py
@@ -70,7 +70,7 @@ class VoxelSelector:
         If raw_data2 is specified, len(raw_data) MUST equal len(raw_data2),
         the correlation will be computed as raw_data by raw_data2.
 
-    voxel_unit: int, default 100
+    voxel_unit: int, default 64
         The number of voxels assigned to a worker each time
 
     master_rank: int, default 0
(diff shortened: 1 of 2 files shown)


brainiak/fcma/voxelselector.py
```diff
@@ -28,6 +28,7 @@
 from . import fcma_extension
 from . import cython_blas as blas
 import logging
+import pathos.multiprocessing
```
mihaic: Given that you do not use other features of Pathos, could you please use the standalone Multiprocess package?
https://pypi.python.org/pypi/multiprocess
We are trying to remove the dependency on Pathos (see issue #149).

yidawang: According to @mjanderson09 , the standalone multiprocessing can only serialize methods at the root level, and the Pathos version is more flexible so that we can define a lambda function and use it in the `map` method.

mihaic: Multiprocess can serialize lambdas, as shown in the last example in its README:
https://github.com/uqfoundation/multiprocess#examples

yidawang: Sorry I didn't make myself clear. It is not just about lambda functions. A lambda function without involving any other user-defined methods should probably be fine, but in my case, I have `_cross_validation_for_one_voxel` used in my lambda function. If I want to use the standalone multiprocessing, I will have to put `_cross_validation_for_one_voxel` outside the class and define it at the root level.

PR #177, opened 2017-02-08
replace pathos.multiprocessing with multiprocess

yidawang:
This is to get rid of the dependency of `pathos`. It reported "Too many open files error" before but is fixed by wrapping the usage of `Pool` using `with`.

### DIFF
--- a/brainiak/fcma/voxelselector.py
+++ b/brainiak/fcma/voxelselector.py
@@ -413,10 +413,8 @@ class VoxelSelector:
                        self.labels) for i in range(task[1])]
 
             with multiprocess.Pool(None) as pool:
-                results = list(pool.map(
-                    lambda x: _cross_validation_for_one_voxel
-                    (x[0], x[1], x[2], x[3]),
-                    inlist))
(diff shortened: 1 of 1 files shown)


brainiak/fcma/voxelselector.py
```diff
@@ -412,11 +412,11 @@ def _cross_validation_for_one_voxel(vid, num_folds,
...
-            pool = pathos.multiprocessing.ProcessingPool(None)
-            results = list(pool.map(
-                lambda x: _cross_validation_for_one_voxel
-                (x[0], x[1], x[2], x[3]),
-                inlist))
+            with multiprocess.Pool(None) as pool:
+                results = list(pool.map(
+                    lambda x: _cross_validation_for_one_voxel
+                    (x[0], x[1], x[2], x[3]),
+                    inlist))
```
mihaic: How about:
```
results = [pool.starmap(_cross_validation_for_one_voxel, inlist)]
```

yidawang: done

PR #181, opened 2017-02-16
Add searchlight srm example

hejiaz:
Add searchlight srm time segment matching example


buildbot-princeton:

buildbot-princeton:

mihaic:
Jenkins, add to whitelist.

examples/funcalign/searchlight_srm_example.py
```diff
@@ -0,0 +1,152 @@
...
+    tst_data = np.zeros((ndim*win_size, nseg))
+    for w in range(win_size):
+      tst_data[w*ndim:(w+1)*ndim,:] = data[tst_subj][:,w:(w+nseg)]
+
+    A =  np.nan_to_num(stats.zscore((trn_data - tst_data),axis=0, ddof=1))
+    B =  np.nan_to_num(stats.zscore(tst_data,axis=0, ddof=1))
+    # normalize A and B
+    A = A/np.linalg.norm(A,axis=0)
+    B = B/np.linalg.norm(B,axis=0)
+    corr_mtx = B.T.dot(A)
```
mihaic: Could you use our optimized function?
http://pythonhosted.org/brainiak/brainiak.fcma.html#brainiak.fcma.util.compute_correlation

examples/funcalign/searchlight_srm_example.py
```diff
@@ -0,0 +1,152 @@
...
+import sys
+import scipy.io as sio
+from scipy.stats import stats
+from brainiak.searchlight.searchlight import Searchlight
+from brainiak.funcalign.srm import SRM
+import warnings
+
+# parameters
+sl_rad = 1 #searchlight length (of each edge) will be 1+2*sl_rad
+max_blk_edge = 1 #won't change computational results, has effect on MPI processing
```
mihaic: You should leave `max_blk_edge` out, unless you want to provide some insight into its impact on performance. Searchlight will use the default value if you do not provide it.

examples/funcalign/searchlight_srm_example.py
```diff
@@ -0,0 +1,152 @@
...
+  # return: can also return several values. In that case, the final output will be 
+  # a 3D array of tuples
+  return np.mean(accu) 
+
+# Run searchlight
+acc = sl.run_searchlight(sfn) # output is a 3D array in shape (dim1,dim2,dim3)
+
+# save result
+if rank == 0:
+  print (acc)
```
mihaic: It would be helpful to print more information while the example is running, so the user understands what is happening.

examples/funcalign/searchlight_srm_example.py
```diff
@@ -0,0 +1,152 @@
...
+  np.savez_compressed('data/searchlight_srm_tsm_acc.npz',acc=acc)  
+
+
+
+
```
mihaic: We do not automatically check the coding style of examples, but it would still be good to follow the recommendations in the contributing instructions:
https://github.com/IntelPNI/brainiak/blob/master/CONTRIBUTING.rst

In particular, the indent should be four spaces, not two, variable and function names should be words connected with underscores, and docstrings should follow the NumPy standard as implemented in Sphinx:

examples/funcalign/searchlight_srm_time_segment_matching_example.ipynb
```diff
@@ -0,0 +1,498 @@
...
+      "  [None None None None None None None None None None]\r\n",
+      "  [None None None None None None None None None None]\r\n",
+      "  [None None None None None None None None None None]\r\n",
+      "  [None None None None None None None None None None]\r\n",
+      "  [None None None None None None None None None None]]]\r\n"
+     ]
+    }
+   ],
+   "source": [
+    "! mpirun -n 16 python3 searchlight_srm_example.py"
```
mihaic: Users running the code on their laptops will be better off with `-n 4`.

examples/funcalign/searchlight_srm_example.py
```diff
@@ -0,0 +1,152 @@
...
+movie_data = movie_file['data']
+
+# Dataset size parameters
+dim1,dim2,dim3,ntr,nsubj = movie_data.shape
+
+# preprocess data, zscore and set NaN to 0
+all_data = [] # first half train, second half test
+for s in range(nsubj):
+  with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
```
mihaic: Removing this warning filter, I see several warnings like this:
`RuntimeWarning: invalid value encountered in true_divide`

Is this what you want to ignore? Is there no way to prevent them, instead of ignoring them?

examples/funcalign/searchlight_srm_time_segment_matching_example.ipynb
```diff
@@ -0,0 +1,498 @@
...
+    "### Run Searchlight SRM"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
```
mihaic: Please use nbstripout to remove the output from notebooks before checking them in.
https://pypi.python.org/pypi/nbstripout

examples/funcalign/searchlight_srm_example.py
hejiaz: Yes, this is the warning I want to ignore. There are some 0 voxels in the data which I have to keep, so I will get the warning above when z-scoring the data. I can z-score voxel by voxel and manually check if it is 0, but it would be slower and more complicated. It should be safe to ignore this warning.

mihaic: Can you please add a comment with this explanation next to the warnings filter?

hejiaz: fixed

mihaic:
Thank you for your contribution, @hejiaz.

PR #182, opened 2017-02-17
[docs] Recommend nbstripout

mihaic:


mihaic:
If we want to present output to users, we can add it to the documentation. The notebooks in the examples are supposed to be run by users; I do not see the need for their output being under version control.

PR #184, opened 2017-02-21
[fmrisim] Fix self mask in example

mihaic:


mihaic:
@CameronTEllis, this is a small change that seems obvious to me, but can you have a look nevertheless?

PR #189, opened 2017-02-27
Fmrisim update

CameronTEllis:


brainiak/utils/fmrisim.py
```diff
@@ -571,6 +579,9 @@ def double_gamma_hrf(stimfunction,
         scale_function : bool
             Do you want to scale the function to a range of 1
 
+        temporal_resolution : float
+          What is the temporal resolution (in s) of the stimulus function.
+          In other words, how many elements per second?
```
mihaic: I think the note taken from the old comment, `(in s)`, is misleading here. Why not have the same description for `temporal_resolution` as in the other functions?

brainiak/utils/fmrisim.py
```diff
@@ -618,11 +627,12 @@ def double_gamma_hrf(stimfunction,
     signal_function = np.convolve(stimfunction, hrf)
 
     # Decimate the signal function so that it only has one element per TR
-    signal_function = signal_function[0::int(tr_duration * 1000)]
+    temp = int(tr_duration * temporal_resolution)
+    signal_function = signal_function[0::temp]
```
mihaic: Please change `temp` to a meaningful name.

brainiak/utils/fmrisim.py
```diff
@@ -443,11 +447,9 @@ def generate_stimfunction(onsets,
...
         weights = weights * len(onsets)
 
-    # How many elements per second are you modeling
-    temporal_resolution = 1000
-
-    # Generate the time course as empty, each element is a millisecond
-    stimfunction = [0] * round(total_time * temporal_resolution)
+    # Generate the time course as empty, each element is a millisecond by
+    # default
+    stimfunction = [0] * round(int(total_time * temporal_resolution))
```
CameronTEllis: I implemented this change because I was getting the type error in Jenkins (which happened again this time). Do you know why it is saying that this is a float? I know that total_time is an, int and temporal_resolution is a float but they seem to play nicely with each other.

mihaic: You bumped into a NumPy issue:
https://github.com/numpy/numpy/issues/5700

I suggest you create an `int` using the result of `round`, which preserves the rounding instead of truncating:
```int(round(total_time * temporal_resolution))```

mihaic: @CameronTEllis, for some reason my previous comment (https://github.com/IntelPNI/brainiak/pull/189#discussion_r103521681) does not show up on Slack, so I am mentioning you to make sure you see it.

PR #191, opened 2017-03-02
allow users to specify the number of processes

yidawang:
Allow users to specify the number of processes to be used in the cross-validation of correlation-based voxel selection. This is to make sure the program wouldn't spawn too many processes to blow up the memory.

Thanks in advance for reviewing.

### DIFF
--- a/brainiak/fcma/voxelselector.py
+++ b/brainiak/fcma/voxelselector.py
@@ -73,6 +73,10 @@ class VoxelSelector:
     voxel_unit: int, default 64
         The number of voxels assigned to a worker each time
 
+    process_num: Optional[int]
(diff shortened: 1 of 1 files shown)


brainiak/fcma/voxelselector.py
```diff
@@ -73,6 +73,10 @@ class VoxelSelector:
     voxel_unit: int, default 64
         The number of voxels assigned to a worker each time
 
+    process_num: Optional[int]
+        The number of processes used in cross validation.
+        If None, it will use up all available cores.
+
```
mihaic: We could use more precise language, e.g., "If None, the number of processes will equal the number of CPU cores."

PR #194, opened 2017-03-04
Add recent work to BRSA

lcnature:
1. Add group BRSA: shared covariance matrix across subjects.
2. Add transform and score function
3. Modify utils.gen_design to make the generated design matrix approximately scaled in amplitude.
4. Add automatic determination of number of nuisance regressor based on method by Gavish and Donoho 2013


brainiak/reprsimil/brsa.py
```diff
@@ -1669,3 +2290,1292 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+                * (- n_run * rho1[i_v] / (1 - rho1[i_v]**2)
+                   - np.einsum('ij,ij', X0TAX0_i[i_v, :, :],
+                               dX0TAX0_drho1[i_v, :, :]) * 0.5
+                   - dYTAcorrY_drho1_ele * 0.5 / sigma2[i_v])
+        deriv = deriv_a1
+
+        return -LL, -deriv
+
+
+class GBRSA(BRSA):
```
mihaic: If any of the docstrings are the same, you can reference BRSA instead of copying the text. This will save time later on when you change the text.

examples/reprsimil/gbrsa_group_representational_similarity_estimate_example.ipynb
```diff
@@ -0,0 +1,907 @@
+{
```
mihaic: Do you mind shortening the name, in line with recommendations for regular Python module names? It could be `gbrsa_example.ipynb`.

If you agree, please also rename the BRSA example.

tests/reprsimil/test_gbrsa.py
```diff
@@ -0,0 +1,331 @@
+#  Copyright 2016 Mingbo Cai, Princeton Neuroscience Instititute,
```
mihaic: The tests in these modules are running too long. This is why the tests are not passing. I know it is difficult, but please shorten the test to less than 1 minute in total for `reprsimil`.

mihaic:
@lcnature, I left some comments, but we need someone with more BRSA experience looking at the code. Could you please nominate someone?

lcnature:
Thanks Mihai! Either Mike Anderson or Mike Shvartsman is good.

mihaic:
@mshvartsman, @mjanderson09, any chance you can have a look at this PR?

brainiak/reprsimil/brsa.py
```diff
@@ -169,24 +228,56 @@ class BRSA(BaseEstimator):
...
+        elif nureg_method == 'PCA':
+            self.nureg_method = PCA(n_components=n_nureg)
+        elif nureg_method == 'SPCA':
+            self.nureg_method = SparsePCA(n_components=n_nureg,
+                                          max_iter=20, tol=tol)
+        elif nureg_method == 'ICA':
+            self.nureg_method = FastICA(n_components=n_nureg, whiten=True)
+        else:
+            raise ValueError('nureg_method can only be FA, PCA, '
+                             'SPCA(for sparse PCA) or ICA')
```
lcnature: Thanks Mihai, I think I will make it able to take both. Part of is because some objects have optional parameters. And I do not want to force the user to look for scipy objects in order to use the method.

brainiak/reprsimil/brsa.py
lcnature: Hi @mihaic Do you mean literally write "refer to BRSA", or some magic way of showing the docstring of BRSA in the docstring of GBRSA (and add new things to it)?

mihaic: There are ways to combine the docstrings such that they are visible via introspection:
https://github.com/meowklaski/custom_inherit

However, I am not convinced we should adopt such a solution, especially given that it is not included in the Python standard library.

Therefore, you should include a Sphinx reference, e.g.:

    See `.BRSA`.

brainiak/reprsimil/brsa.py
mihaic: Why do you care what object the user passes in, as long as it is compatible with scikit-learn, so you can call `fit_transform`?

Furthermore, given that you provide a default, the user is not forced to look elsewhere. Of course, users wanting a different analysis should understand what that is, so I think it is reasonable to have them instantiate a scikit-learn object.

lcnature: OK. Then I will keep both. If they send anything else other than a string, I won't complain. But if they send any string other than the few suggested, then I will give an error.

mihaic: I brought this up in the first place to remove the string argument. It makes it difficult to reason about the code, because you must understand the `if` statement that analyzes the string.

If you are happy with the scenario I described, where the user instantiates and passes an object in case they do not want the default method, why not eliminate the string option? Conversely, there is no point in making the API more complex by adding the object option if you need the string option.

lcnature: I see. Part of the reason to keep the string option is in future I want to test using (h)tfa or srm in place of PCA/ICA/FA. Those might take list of subjects as input instead of single numpy array. So the code needs to deal with them differently from other objects that takes simply numpy arrays. If indeed I implement this in future, then only accepting an object seems difficult to deal with because the code does not know whether the object the user feeds in take single subject or multiple subjects. Of course  it can be done by adding an extra parameter which indicates whether the object is for single subject or multiple subjects, but that requires the user to make sure that they
put the correct value for that parameter together with the object they feed in.
I just think that keeping the option a string makes it simple for user. But I do like the idea of making it able to accept an object like scipy's minimizer does.

lcnature: If you are concerned with the `if` statement, how about I use a dictionary to replace it?

mihaic: I understand your need for string. The `if` statement is not the problem; the problem is having additional logic deciding what the function. However, because the scikit-learn API does not cover the functionality wee need (e.g., a list of subjects as input), the extra logic is unavoidable. So leave the code as it is.

brainiak/reprsimil/brsa.py
lcnature: Hi @mihaic I realized that there is still some difference between the two docstring. So I only added the reference without removing the basic description of GBRSA.

lcnature:
@mihaic I cannot really see which line has wrong indentation from the current report. Can you help on that?

mihaic:
@lcnature, see https://github.com/lcnature/brainiak/pull/2.

brainiak/reprsimil/brsa.py
```diff
@@ -274,9 +275,11 @@ def __init__(
                                           max_iter=20, tol=tol)
         elif nureg_method == 'ICA':
             self.nureg_method = FastICA(n_components=n_nureg, whiten=True)
-        else:
+        elif type(nureg_method) is str:
             raise ValueError('nureg_method can only be FA, PCA, '
                              'SPCA(for sparse PCA) or ICA')
+        else:
+            self.nureg_method = nureg_method
```
mihaic: I think accepting strings and instantiated objects is worse than only accepting strings, because it makes reasoning about the code more complicated. Please revert to the initial code that only accepts strings.

brainiak/reprsimil/brsa.py
```diff
@@ -891,6 +917,29 @@ def _calc_dist2_GP(self, coords=None, inten=None,
...
+            log_ptau = scipy.stats.halfcauchy.logpdf(
+                tau2**0.5, scale=self.tau_range)
+        else:
+            tau2 = (log_SNR_invK_tilde_log_SNR + 2 * self.tau_range**2) /\
+                (2 * 2 + 2 + n_V)
+            log_ptau = scipy.stats.invgamma.logpdf(
+                tau2, scale=self.tau_range**2, a=2)
+        # LL_inc is  p(tau2)
+        return tau2, log_ptau
+
```
mihaic: My original comment went further than this refactoring. I was suggesting creating two functions, e.g., `estimate_std_dev_gp_log_snr_half_cauchy` and `estimate_std_dev_gp_log_snr_inv_gamma`, and having the user pass one of the functions instead of a the `tau2_prior` string to `BRSA`.

tests/reprsimil/test_gbrsa.py
```diff
@@ -168,24 +168,24 @@ def test_fit():
...
+    # assert p < 0.01, "Fitted covariance matrix does not correlate with ideal covariance matrix!"
+    # # check that the recovered SNRs makes sense
+    # p = scipy.stats.pearsonr(gbrsa.nSNR_[1],snr[1])[1]
+    # assert p < 0.01, "Fitted SNR does not correlate with simulated SNR!"
+    # p = scipy.stats.pearsonr(gbrsa.sigma_[2],noise_level[2])[1]
+    # assert p < 0.01, "Fitted noise level does not correlate with simulated noise level!"
+    # p = scipy.stats.pearsonr(gbrsa.rho_[0],rho1[0])[1]
+    # assert p < 0.01, "Fitted AR(1) coefficient does not correlate with simulated values!"
+    # p = scipy.stats.pearsonr(gbrsa.beta0_[0][0, :],inten[0])[1]
+    # assert p < 0.01, "Fitted intensity level does not correlate with the simulated value!"
```
mihaic: Remove this code if you are not using it.

brainiak/reprsimil/brsa.py
lcnature: Sorry I did not understand this. Do you mean making these as functions out of BRSA object? How can the user know there exists such an object? It seems easier to use by just choosing one of the two strings.

mihaic: Exactly. You document their usage just like you document the usage of the current strings, except that the code now is easy to inspect and understand. Furthermore, it is easy to add new functionality by adding a new function, instead of adding code to a single big function that behaves according to a string parameter.

mihaic: Think about the scikit-learn API, for example. Instead of using the string "SVM" to select the classification method, you actually pass an `SVC` object.

lcnature: Hi @mihaic it was not clear to me how to write the __init__ in this case. If I make those two functions within brsa module but outside BRSA class, then I get error saying E               NameError: name 'estimate_var_gp_log_snr_half_cauchy' is not defined

mihaic: That is because in Python "default values are evaluated at the point of function definition":
https://docs.python.org/3/tutorial/controlflow.html#default-argument-values

Therefore, the function definition must be placed before the class definition in the module for the function name to be usable as the default argument value in `__init__`.

lcnature:
Thanks!
Now as I think of it, I am still worried that asking users to choose from
two functions that they do not know might deter some people. "invGamma" is
still something interpretable, at least they know this refers to a form of
distribution. But having to choose from two functions of long names seem
quite scary. How about I make two functions either within or outside BRSA
class, but direct a function of _calc_tau2 to one of them based on the
string argument? If advanced user wants to try a different form of prior
distribution, they can still add one and make a pull request.
I somehow still prefer the possibility of accepting either a string or a
function object, such that for most users, the string argument is easy
enough to guess what it means, and for advanced users, they can write their
own function as long as it has the same interface.

mihaic:
@lcnature, this is the same discussion we had for `nureg_method`. Whatever intuition users can have reading the string "invGamma", they can also have reading the function name, which can be whatever you (as the module designer) choose, including "estimate_inv_gamma". Is that not as intuitive?

lcnature:
@mihaic I don't know. I just have a feeling that for users relatively new to Python, this might create another confusion. I moved the two functions to the top of brsa.py and it worked. But I see that when I hit tab+shift, the usage instruction shown for BRSA shows the default argument like this "tau2_prior=<function prior_GP_std_halfCauchy at 0x7f36ff593ae8>" Think of someone who is used to Matlab, this might appear quite strange and him/her might not guess out how to replace it with another function, and where to find it.
It is true that the user can always find another function within brsa module, or we can point them to try the other prior function in the module. But wouldn't it be complicated that they have to write BRSA(..., tau2_prior=brsa.prior_GP_std_invGamma,...)? Also sometimes user might just do `from brsa import BRSA`, then it is hard for them to see the extra function within brsa. This function is only used within BRSA, not like `svm` that can be widely used.

mihaic:
@lcnature, you can address the issues you noticed.

You could name the argument, `tau2_prior_fn`, to suggest the casual user that it holds a function.

Regarding instructions for alternatives, you should point users to `prior_GP_std_inv_gamma` in the docstring of `tau2_prior_fn`, where you document the possible string values now. (Note that you should not add "camelCase" to function names.)

Finally, for the import problem, you can make the functions static function in `BRSA`, so they are accessible without further imports.

lcnature:
Hi @mihaic I am trying to define it as static method.
So now I place 
`@staticmethod
    def _prior_GP_std_half_cauchy(log_SNR_invK_tilde_log_SNR, n_V, tau_range):
        tau2 = (log_SNR_invK_tilde_log_SNR - n_V * tau_range**2
                + np.sqrt(n_V**2 * tau_range**4 + (2 * n_V + 8)
                          * tau_range**2
                          * log_SNR_invK_tilde_log_SNR
                          + log_SNR_invK_tilde_log_SNR**2))\
            / 2 / (n_V + 2)
        log_ptau = scipy.stats.halfcauchy.logpdf(
            tau2**0.5, scale=tau_range)
        return log_ptau, tau2`
within BRSA class. But how should I write the default argument in `__init__`? If I just write `tau2_prior=_prior_GP_std_half_cauchy`, python complains that the function is not defined. If I write `tau2_prior=self._prior_GP_std_half_cauchy` or `tau2_prior=BRSA._prior_GP_std_half_cauchy`, then neither self or BRSA is defined. Can you explain how to use a static method within the class?

mihaic:
@lcnature, you can do one of the following:
* Use `None` as default value and set the actual value in the body of `__init__`.
* Move the definition before `__init__` and access the function wrapped by the method as `prior_GP_std_half_cauchy.__func__` (see https://docs.python.org/3/reference/datamodel.html)
* Move the functions to the module level (users would still find the alternative function if you reference it in the parameter docstring).

Also, the names must be public to be accessible to users, so remove the leading `_`.

brainiak/reprsimil/brsa.py
```diff
@@ -30,24 +32,137 @@
...
+        of the inverse-Gamma prior.
+        The smaller it is, the more penalization is imposed
+        on large variation of y.
+    Returns:
+    --------
+    tau2: The MAP estimation of tau^2 based on the prior on tau
+        and y_invK_y.
+    log_ptau: log(p(tau)) of the returned tau^2 based on the
+        inverse-Gamma prior.
+    """
```
mihaic: You should not duplicate the docstring of `prior_GP_var_half_cauchy`. Say what the function does in a one-line summary, add a sentence on using inverse gamma, and then say "See \`.prior_GP_var_half_cauchy\` for details."

lcnature: OK. Done.

lcnature: BTW, I think the block literal did not do anything to the docstring. It simply displayed the text as is without any special formatting. Can you have a look at the docstring of BRSA and look for the part for tau2_prior, and let me know how it should be written instead?

mshvartsman:
@mihaic, let me know when you are happy with the PR from a code quality standpoint and then I'll work through the math / science side of things. @lcnature, do you have a scientific description of the changes written up somewhere? I have a hunch of how GBRSA should work but it would be nice to have math to go with the code.

lcnature:
@mshvartsman Thanks! I have some notes written myself but not in a polished way. Still writing the paper.

mihaic:
@mshvartsman, thanks for helping. The code looks good. There will probably be a few more minor changes, but nothing major.

brainiak/reprsimil/brsa.py
mihaic: It did not do anything because it is not a literal block. Have another look at the definition of literal blocks:
http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#literal-blocks
and the Sphinx NumPy-style example:

You need two colons (`::`), indentation, and blank lines separating the example from the surrounding text.

brainiak/reprsimil/brsa.py
```diff
@@ -87,106 +176,205 @@ class BRSA(BaseEstimator):
...
+        while half-Cauchy prior only penalizes for very large tau.
+        Both the functions are part of brsa module, which you need
+        to import together with BRSA in order to use.
+        For example::
+        from brainiak.reprsimil.brsa import BRSA, prior_GP_var_inv_gamma
+        brsa = BRSA(tau2_prior=prior_GP_var_inv_gamma)
+        Or, if you do are happy with the default inverse-Gamma prior,
+        then you can ignore this argument::
+        from brainiak.reprsimil.brsa import BRSA
+        brsa = BRSA()
```
mihaic: I suggest you keep only this usage example and remove the examples from the two functions.

Furthermore, you should remove the descriptions of the functions, which are already in the function docstrings, where they belong. Make it easier for users to jump to the functions in the HTML docs by marking the functions (e.g., `` `.prior_GP_var_half_cauchy` ``).

brainiak/reprsimil/brsa.py
```diff
@@ -254,20 +252,22 @@ class BRSA(BaseEstimator, TransformerMixin):
...
+
+            from brainiak.reprsimil.brsa import BRSA
+            brsa = BRSA()
+
+        If you want to try the alternative half-Cauchy prior,
+        then you need to import it in addition to BRSA::
+
+            from brainiak.reprsimil.brsa import BRSA, prior_GP_var_half_cauchy
+            brsa = BRSA(tau2_prior=prior_GP_var_half_cauchy)
+
```
mihaic: You must also add the functions to `__all__` to make them part of the API. This will also add them to the HTML docs.

mihaic:
@mshvartsman, @lcnature, any update?

mihaic:
@lcnature, how is this PR going?

mshvartsman:
I'm the bottleneck at the moment -- but I'm actually slightly dug out so I will try to get through it this week.

mshvartsman:
I recommend making GBRSA and BRSA in separate files `brsa.py` and `gbrsa.py`. They're distinct methods and I think it should be a bit easier to keep track of what's where if they are distinct (see e.g. srm.py and sssrm.py).

lcnature:
@mihaic The suggestion by @mshvartsman is fine with me. But I wonder what is your suggestion for code organization and user friendliness. FYI, in the long run, I would recommend users to use gbrsa for most usages. But not knowing what disadvantage it might have, I will keep both brsa and gbrsa in place.

mihaic:
If BRSA and GBRSA have the same objective and share code, they should be in the same file, even if this makes the file unusually long.

mshvartsman:
Hey @cbaldassano, do you mind reviewing just the forward-backward algo in here? I haven't touched HMMs in years and it's probably fresher in your mind than in mine.

cbaldassano:
The overall logic of _forward_step and _backward_step looks good, though it is hard for me to evaluate the math without seeing a writeup of what it is supposed to be computing.

The code could use some comments, and there are some lines that are commented out (e.g. 1529, 1533-1534) that are unclear if they are important.

lcnature:
Thanks @cbaldassano ! I will send you a pdf of some details of the derivation. And I will add more comments to the code and clean up those unused codes.

brainiak/reprsimil/brsa.py
```diff
@@ -344,34 +617,170 @@ def fit(self, X, design, nuisance=None, scan_onsets=None, coords=None,
...
+            in the design matrix of the training data.
+            Unknown nuisance time series will be marginalized, assuming
+            they follow the same spatial pattern as in the training
+            data. The hypothetical response captured by the design matrix
+            will be subtracted from data before the marginalization
+            when evaluating the log likelihood. For null model,
+            nothing will be subtracted before marginalization.
+            If you z-scored your data during fit step, you should
+            z-score them for score function as well. If you did not
+            z-score in fitting, you should not z-score here either.
```
mihaic: Right. You could create a boolean `z_score` class parameter.

brainiak/reprsimil/brsa.py
```diff
@@ -30,24 +32,158 @@
...
 
 
-class BRSA(BaseEstimator):
+def prior_GP_var_inv_gamma(y_invK_y, n_y, tau_range):
+    """ Imposing a prior onto the variance (tau^2) of the Gaussian
+        Process which is in turn a prior imposed over
+        a function y = f(x).
+        This function imposes an inverse-Gamma prior on tau^2
+        with a shape parameter alpha=2 and rate parameter
+        beta=tau_range^2.
```
lcnature: Sorry. It should be "scale".

brainiak/reprsimil/brsa.py
```diff
@@ -30,24 +32,158 @@
...
+        The function returns the MAP estimate of tau^2 and
+        log(p(tau^2|tau_range)) for the estimated value of tau^2,
+        where tau_range describes the reasonable range of tau
```
lcnature: Unfortunately there is not such an easy mapping. So I may not make such extrapolation.

brainiak/reprsimil/brsa.py
```diff
@@ -30,24 +32,158 @@
...
+
+
+def Ncomp_BIC_Minka(X):
+    """ This implements the BIC metric written in Minka 2000:
+        Automatic choice of dimensionality for PCA
+        The paper mainly proposed a Laplace approximation of the posterior
+        of the number of components but also provided this BIC approximation.
+        For cases in which feature number is smaller than sample size,
+        there is ambiguity for the author of this code to decide how to
+        set some variables. Empirically the number of components estimated
```
lcnature: Sorry about the confusion. If n is the number of observations, and k the number of dimensions, the paper requires k to be smaller than n. 
So the full story is: vanilla implementation of the method often decides to choose k equal to the maximal possible value (rank of the data matrix - 1) for real fMRI data. The approximate log(p(D|M)) given by BIC often appears as a shape of "2" rotated by 90 degrees, with the right-hand tail higher than the peak on the left. This is undesirable and strange. My interpretation is that the BIC approximation works when the number of observation is sufficiently large (as in https://en.wikipedia.org/wiki/Bayesian_information_criterion, although I do not know the correct criterion of calling n "large"). For fMRI data which often have more channels (voxels) than observations, the penalization term might not be sufficient. With a lot of trial-and-error, using larger penalization (n_T * n_V) as the number of observations manages to suppress the tail. But the chosen k is often much smaller than Ncomp_SVHT_MG_DLD_approx gives. Simulation result also shows that the k chosen by Ncomp_SVHT_MG_DLD_approx gives better performance in terms of the recovered similarity matrix.

So, I think I will just remove this BIC metric. I might give a try of the formula not using BIC approximation in the paper. If it fails, Ncomp_SVHT_MG_DLD_approx will be the only option.

brainiak/reprsimil/brsa.py
```diff
@@ -30,24 +32,158 @@
...
+    m = Lambda.size * K - K * (K + 1) / 2
+    v_hat = (np.sum(Lambda) - np.cumsum(Lambda[:d_max])) / (Lambda.size - K)
+    log_Lambda = np.log(Lambda)
+    log_p = - n_T / 2 * np.cumsum(log_Lambda[:d_max]) \
+        - n_T * (Lambda.size - K) / 2 * np.log(v_hat) \
+        - (m + K) / 2 * np.log(n_T * n_V)
+    # Equation (78) of Minka 2000. Notice that we operate on the
+    # space spanned by the eigen vectors. Although Minka counted the
+    # number of data point with n_T in this case, we count observations
+    # of all time points and all voxels (n_T * n_V)
```
lcnature: No mathematical justification. See the comment above. Yeah I know this is super hacky.

brainiak/reprsimil/brsa.py
```diff
@@ -30,24 +32,158 @@
...
+    ncomp = np.argmax(log_p) + 1
+    return ncomp, log_p
+
+
+class BRSA(BaseEstimator, TransformerMixin):
```
lcnature: The line below says "A correlation matrix converted from the covariance matrix
    will be provided as a quantification of neural representational similarity." I will add the notation "U" in the description in reference to the equation below.

brainiak/reprsimil/brsa.py
```diff
@@ -867,38 +1456,264 @@ def _fit_RSA_UV(self, X, Y, X0,
...
+    def _forward_step(self, Y, T_X, Var_X, Var_dX, rho_e, sigma2_e, weight):
+        """ forward step for HMM """
+        # cov_form='diag'  We currently only implement diagonal form
+        # of covariance matrix for Var_X, Var_dX and T_X, which means
+        # each dimension of X is independent and their refreshing noise
+        # are also independent. Note that log_p_data takes this assumption.
+        if Var_X.ndim == 1:
+            inv_Var_X = np.diag(1 / Var_X)
+            half_log_det_Var_X = np.sum(np.log(Var_X)) / 2.0
+            Var_X = np.diag(Var_X)
```
cbaldassano: Will this line (and similar assignments to Var_x and Var_dX below) cause changes to the input variables outside the scope of this function?

cbaldassano:
I took another look at the forward-backward functions now that I've seen the math, and it looks like they line up correctly. I just made a minor comment about possible modification of the input arguments.

brainiak/reprsimil/brsa.py
mihaic: It will not. However, there is also the danger of modifying mutable objects; hopefully that is not happening here.
https://docs.python.org/3/faq/programming.html#how-do-i-write-a-function-with-output-parameters-call-by-reference

brainiak/reprsimil/brsa.py
```diff
@@ -30,24 +32,158 @@
...
+        there is ambiguity for the author of this code to decide how to
+        set some variables. Empirically the number of components estimated
+        turn to be very large by the Laplace approximation method. The BIC
+        method appear to perform well in a limited number of datasets.
+    Parameters
+    ----------
+    X: 2-D numpy array of size [n_T, n_V]
+        The data to estimate the dimensionality for PCA.
+        Each column is one feature. Each row is one sample.
+        X is z-scored before further calculation.
```
lcnature: That makes sense. The model right now does not restrict the data to be z-scored. And this step probably does not take too much time. So perhaps we can leave it as is to reduce the number of options?

mshvartsman: In general that would be fine -- but you have instructions in a few other docstrings to "zscore here the same way you zscore elsewhere" -- so I would say if you always zscore, just always zscore everywhere. I still think letting the user handle the z-score or passing as an option is better.

brainiak/reprsimil/brsa.py
```diff
@@ -62,22 +198,26 @@ class BRSA(BaseEstimator):
...
-    n_iter : int, default: 200
+    n_iter : int. Default: 50
         Number of maximum iterations to run the algorithm.
-    rank : int, default: None
+    n_iter_inner : int, default: 6
+        Number of maximum iterations in the inner loop of optimization (within
+        each iteration of n_iter). Users typically do not need to adjust this
+        parameter.
+    rank : int. Default: None
         The rank of the covariance matrix.
```
lcnature: It is a degenerate likelihood. I did some reparametrization: beta = (s*sigma) * L * alpha. alpha ~ N(0, I). To make the covariance matrix U low rank, we basically restrict the number of columns of L to be the rank chosen by the user. Then alpha is also lower dimensional.

brainiak/reprsimil/brsa.py
```diff
@@ -87,106 +227,220 @@ class BRSA(BaseEstimator):
...
+    n_nureg: int or string. Default: 'BIC'
         Number of nuisance regressors to use in order to model signals
         shared across voxels not captured by the design matrix.
-        This parameter will not be effective in the first round of fitting.
-    GP_space: boolean, default: False
+        This number is in addition to any nuisance regressor that the user
+        has already provided.
+        If set to 'BIC', the number of nuisance regressor will be
+        automatically determined based on BIC metric
+        of probabalistic PCA model. (Minka 2000, NIPS)
```
lcnature: This function will be removed.

brainiak/reprsimil/brsa.py
```diff
@@ -87,106 +227,220 @@ class BRSA(BaseEstimator):
...
+    nureg_method: string, 'PCA', 'FA', 'ICA' or 'SPCA'. Default: 'ICA'
```
lcnature: For any way of choosing number of components, say the one we keep now, the number makes sense for ICA because PCA is usually the first step of ICA. The assumption for PCA might not directly apply to FA. I am not sure about SPCA. But practically I find the result of SPCA is very close to PCA. I do not know if there are other ways to determine the number of components for FA. But I want to leave this option available in case people want to try.

brainiak/reprsimil/brsa.py
```diff
@@ -87,106 +227,220 @@ class BRSA(BaseEstimator):
...
+        The method to estimate the shared component in noise across voxels.
+    DC_single: boolean, default: True
```
lcnature: Thanks! I think both terms are equally good to me.

brainiak/reprsimil/brsa.py
```diff
@@ -87,106 +227,220 @@ class BRSA(BaseEstimator):
...
+        believe the average image intensity might not scale with the
+        same proportion for different voxels across scan. In other words,
+        it is possible that some part of the brain is more vulnerable to
+        change in baseline intensity due to facts such as
+        field inhomogeneity. Setting DC_single to True will force the
+        nuisance regressors automatically estimated from residuals to
+        capture this.
+    GP_space: boolean. Default: False
         Whether to impose a Gaussion Process (GP) prior on the log(pseudo-SNR).
         If true, the GP has a kernel defined over spatial coordinate.
```
lcnature: Thanks. I will make it clear that it is the SNR that is assumed to be smooth.

brainiak/reprsimil/brsa.py
```diff
@@ -87,106 +227,220 @@ class BRSA(BaseEstimator):
...
+        How many initial iterations to fit the model
         without introducing the GP prior before fitting with it,
         if GP_space or GP_inten is requested. This initial
         fitting is to give the parameters a good starting point.
-    optimizer: the optimizer to use for minimizing cost function.
-        We use 'BFGS' as a default. Users can try other optimizer
-        coming with scipy.optimize.minimize, or a custom
-        optimizer.
-    rand_seed : int, default: 0
+    optimizer: str or callable. Default: 'BFGS'
```
lcnature: Good point. I will test it more. Maybe I will change it in future if it is stable across many tests?
I will add the suggestion for optimizer.

brainiak/reprsimil/brsa.py
```diff
@@ -87,106 +227,220 @@ class BRSA(BaseEstimator):
...
-        The pseuso-SNR of all voxels.
+    U_ : numpy array, shape=[condition,condition].
+        The shared covariance matrix.
+    L_ : numpy array, shape=[condition,rank].
+        The Cholesky factor of the shared covariance matrix
+        (lower-triangular matrix).
+    C_: numpy array, shape=[condition,condition].
+        The correlation matrix derived from the shared covariance matrix,
+    nSNR_ : numpy array, shape=[voxels,].
+        The normalized pseuso-SNR of all voxels.
```
lcnature: Sorry, what do you mean here?

brainiak/reprsimil/brsa.py
```diff
@@ -212,38 +468,42 @@ def fit(self, X, design, nuisance=None, scan_onsets=None, coords=None,
...
             If you have multiple scans of the same participants that you
             want to analyze together, you should concatenate them along
             the time dimension after proper preprocessing (e.g. spatial
             alignment), and specify the onsets of each scan in scan_onsets.
-        design: 2-D numpy array, shape=[time_points, conditions]
+        design: numpy array, shape=[time_points, conditions]
             This is the design matrix. It should only include the hypothetic
-            response for task conditions. You do not need to include
+            response for task conditions. You should not include
             regressors for a DC component or motion parameters, unless with
```
lcnature: Good point

brainiak/reprsimil/brsa.py
```diff
@@ -344,34 +617,170 @@ def fit(self, X, design, nuisance=None, scan_onsets=None, coords=None,
...
-                self.sigma_, self.rho_, self.lGPspace_, self.bGP_,\
-                self.lGPinten_ = \
-                self._fit_RSA_UV(X=design, Y=X, X0=nuisance,
+                self._beta_latent_, self.sigma_, self.rho_, \
+                self.lGPspace_, self.bGP_, self.lGPinten_, self.X0_ = \
+                self._fit_RSA_UV(X=design, Y=X, X_base=nuisance,
                                  scan_onsets=scan_onsets,
                                  coords=coords, inten=inten)
 
         self.C_ = utils.cov2corr(self.U_)
```
lcnature: Good point!

brainiak/reprsimil/brsa.py
```diff
@@ -344,34 +617,170 @@ def fit(self, X, design, nuisance=None, scan_onsets=None, coords=None,
...
+            (recommended) when fitting the model, data should be z-scored
+            as well when calling transform()
+        y : not used (as it is unsupervised learning)
+        scan_onsets : numpy array, shape=[number of runs]. Default: None.
+            A list of indices corresponding to the onsets of
+            scans in the data X. If not provided, data will be assumed
+            to be acquired in a continuous scan.
+        Returns
+        -------
+        ts : numpy arrays, shape = [time_points, condition]
```
lcnature: Yes!

brainiak/reprsimil/brsa.py
```diff
@@ -344,34 +617,170 @@ def fit(self, X, design, nuisance=None, scan_onsets=None, coords=None,
...
+        ll = self._score(Y=X, design=design, beta=self.beta_,
+                         scan_onsets=scan_onsets, beta0=self.beta0_,
+                         rho_e=self.rho_, sigma_e=self.sigma_,
+                         rho_X0=self._rho_X0_, sigma2_X0=self._sigma2_X0_)
+        ll_null = self._score(Y=X, design=None, beta=None,
+                              scan_onsets=scan_onsets, beta0=self.beta0_,
+                              rho_e=self.rho_, sigma_e=self.sigma_,
+                              rho_X0=self._rho_X0_,
+                              sigma2_X0=self._sigma2_X0_)
+        return ll, ll_null
```
lcnature: I do not know much of this. I will ask you for more details! Right now this score is for cross-validate data. I think what you proposed would be useful for the case when cross-validation is not possible. Maybe I can do it in the next PR?

brainiak/reprsimil/brsa.py
```diff
@@ -464,51 +873,78 @@ def _prepare_data_XY(self, X, Y, D, F):
...
                         ' The final covariance matrix won''t '
-                        'reflet them.')
-        n_base = X0.shape[1]
-        X0TX0, X0TDX0, X0TFX0 = self._make_templates(D, F, X0, X0)
-        XTX0, XTDX0, XTFX0 = self._make_templates(D, F, X, X0)
-        X0TY, X0TDY, X0TFY = self._make_templates(D, F, X0, Y)
-
-        return X0TX0, X0TDX0, X0TFX0, XTX0, XTDX0, XTFX0, \
-            X0TY, X0TDY, X0TFY, X0, n_base
+                        'reflet these components.')
```
lcnature: Does "won't" work? It is within a string, so I thought won''t works as a escape for special character.

brainiak/reprsimil/brsa.py
```diff
@@ -767,23 +1203,24 @@ def _fit_RSA_UV(self, X, Y, X0,
...
                     XTY, XTDY, XTFY, YTY_diag, YTDY_diag, YTFY_diag,
-                    XTX, XTDX, XTFX, X, Y, X0, D, F, run_TRs,
+                    XTX, XTDX, XTFX, X, Y, X_base, X_res, D, F, run_TRs,
                     current_vec_U_chlsk_l,
                     current_a1, current_logSNR2,
                     idx_param_fitU, idx_param_fitV,
-                    l_idx, n_C, n_T, n_V, n_l, n_run, n_base, rank)
+                    l_idx, n_C, n_T, n_V, n_l, n_run, n_X0, rank)
 
             current_GP[0] = np.log(np.min(
```
lcnature: The length does vary depending on the GP options. But the values always mean the same thing: the 0-th is always the length scale in voxel distance, the 1-st is always the length scale in intensity difference. I think I did this just to save the possible if/else clauses. Otherwise, I need to write separate lines for whether GP_inten is set to True in many places.

brainiak/reprsimil/brsa.py
```diff
@@ -867,38 +1324,258 @@ def _fit_RSA_UV(self, X, Y, X0,
...
+            est_beta_AR1_UV, est_beta0_AR1_UV, est_beta_AR1_UV_latent,\
+            est_sigma_AR1_UV, est_rho1_AR1_UV, est_space_smooth_r, \
+            est_std_log_SNR, est_intensity_kernel_r, X0
+
+    def _transform(self, Y, scan_onsets, beta, beta0,
+                   rho_e, sigma_e, rho_X, sigma2_X, rho_X0, sigma2_X0):
+        """ Given the data Y and the response amplitudes beta and beta0
+            estimated in the fit step, estimate the corresponding X and X0.
+            It is done by a forward-backward algorithm.
+            We assume X and X0 both are vector autoregressive (VAR)
```
lcnature: The reason is practically the intrinsic fluctuation in fMRI and the design matrix are often smooth. For the purpose of decoding, estimating the AR(1) coefficients of X and X0 obtained in training data and use it (together with the variance parameter) as prior can reduce variance in the transforming step. I do not know how to put AR(1) constraint on X0 in fitting step. But they usually come out with autocorrelation automatically if the data has this property. Because X0 is estimated from residuals of so many voxels, I think the data itself provide sufficient constraint without additionally imposing prior during fitting.

brainiak/reprsimil/brsa.py
```diff
@@ -867,38 +1324,258 @@ def _fit_RSA_UV(self, X, Y, X0,
...
+        X0 = np.concatenate(X0)
+        return X, X0, total_log_p
+
+    def _score(self, Y, design, beta, scan_onsets, beta0, rho_e, sigma_e,
+               rho_X0, sigma2_X0):
+        """ Given the data Y, and the spatial pattern beta0
+            of nuisance time series, return the cross-validated score
+            of the data Y given all parameters of the subject estimated
+            during the fist step.
+            The hypothetic response to the task will be subtracted, and
```
lcnature: Since score() is what users actually call instead of _score(), I added explanation of the difference to score()

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2460,1350 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+    The differences of this tool from BRSA are:
+    (1) It allows fitting a shared covariance matrix (which can be converted
+    to similarity matrix) across multiple subjects.
+    This is analogous to SRM under funcalign submodule. Because of using
+    multiple subjects, the result is less noisy.
+    (2) In the fitting process, the SNR and noise parameters are marginalized
+    for each voxel. Therefore, this tool should be faster than BRSA
+    when analyzing an ROI of hundreds to thousands voxels. It does not
+    provide a spatial smoothness prior on SNR though.
+    Both tools provide estimation of SNR and noise parameters at the end,
```
lcnature: Thanks for asking. Those by GBRSA are posterior mean instead of MAP. I will add this to the document.

mihaic:
It's great to see all the activity in this PR. Thank you very much for your reviews, @mshvartsman and @cbaldassano!

@lcnature, given the extensive comments from @mshvartsman, I think it would be helpful for you to actually make commits addressing the comments as you reply to the comments, not later. GitHub does not provide a formal way to mark each comment as resolved, but it would be very helpful leave a message such as "Done" for each comment that is addressed.

brainiak/reprsimil/brsa.py
lcnature: For BRSA, L-BFGS-B seem to not converge close to the true simulated values within the same number of iteration. But it does run faster for the same number of iteration. I changed the default of GBRSA to L-BFGS-B now.

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2460,1350 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+            in nuisance regressors.
+            If you have multiple run, the design matrix
+            of all runs should be concatenated along the time dimension for
+            each participant, with every column for one condition across runs.
+            If the design matrix is the same for all subjects,
+            either provide a list as required, or provide single numpy array.
+        nuisance: optional, list of numpy arrays,
+            shape=[time_points, nuisance_factors] for each subject in the list.
+            Nuisance regressors of each participant.
+            The responses to these regressors will be marginalized out from
```
lcnature: Yes. The variance of the voxel-independent noise is explained by sigma^2. pseudo-SNR and sigma together determines the variance of the task-related signal amplitude beta. But the nuisance regressor variance is not controlled by pseudo-SNR.

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2460,1350 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+                                              scale=self.logS_range)
+        SNR_weights[-1] = 1 - scipy.stats.norm.cdf(log_SNR_grids_upper[-2],
+                                                   scale=self.logS_range)
+
+        # SNR_grids = np.linspace(0, 1, self.SNR_bins)
+        # SNR_weights = np.ones(self.SNR_bins) / self.SNR_bins
+        logger.info('The grids of pseudo-SNR used for numerical integration '
+                    'is {}.'.format(SNR_grids))
+        assert np.isclose(np.sum(SNR_weights), 1), \
+            'The weights for log SNR do not sum to 1!'
```
lcnature: These are now moved to test_gbrsa.py instead

brainiak/reprsimil/brsa.py
lcnature: Thanks! I think it is safe here since Var_X and Var_dX are passed as numpy arrays instead of lists.

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2545,1446 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+
+    def _calc_sandwidge_marginalized(
+            self, XTY, XTDY, XTFY, YTY_diag, YTDY_diag, YTFY_diag,
+            XTX, XTDX, XTFX, X0TX0, X0TDX0, X0TFX0,
+            XTX0, XTDX0, XTFX0, X0TY, X0TDY, X0TFY,
+            rho1, n_V, n_X0):
+        # Calculate the sandwidge terms which put Acorr between X, Y and X0
+        # These terms are used a lot in the likelihood. This function
+        # is used for the marginalized version.
+        XTAY = XTY - rho1[:, None, None] * XTDY \
```
lcnature: solved

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2545,1446 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+                    sigma2_X0=self._sigma2_X0_[subj])
+                ll_null[subj] = self._score(
+                    Y=X[subj], design=None, beta=None,
+                    scan_onsets=scan_onsets[subj], beta0=self.beta0_[subj],
+                    rho_e=self.rho_[subj], sigma_e=self.sigma_[subj],
+                    rho_X0=self._rho_X0_[subj],
+                    sigma2_X0=self._sigma2_X0_[subj])
+        return ll, ll_null
+
+    def _calc_sandwidge_marginalized(
```
lcnature: solved

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2545,1446 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+        # may be modified inside our code.
+
+        scan_onsets = self._check_scan_onsets_GBRSA(scan_onsets, X)
+        # Run Marginalized Bayesian RSA
+        # Note that we have a change of notation here.
+        # Within _fit_RSA_marginalized, design matrix is named X
+        # and data is named Y, to reflect the
+        # generative model that data Y is generated by mixing the response
+        # X to experiment conditions and other neural activity.
+        # However, in fit(), we keep the tradition of scikit-learn that
```
lcnature: fixed

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2460,1350 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+        The normalized pseuso-SNR of all voxels.
+        They are normalized such that the geometric mean is 1
+    sigma_ : list of numpy arrays, shape=[voxels,] for each subject.
+        The estimated standard deviation of the noise in each voxel
+        Assuming AR(1) model, this means the standard deviation
+        of the refreshing noise.
+    rho_ : list of numpy arrays, shape=[voxels,] for each subject.
+        The estimated autoregressive coefficient of each voxel
+    beta_: list of numpy arrays, shape=[conditions, voxels] for each subject.
+        The maximum a posterior estimation of the response amplitudes
```
lcnature: fixed

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2460,1350 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+        We use 1.0 as default. If it is set too small,
+        the estimated SNR will turn to be too close to each other.
+        The code will in fact evaluate SNR in the range of
+        exp(-3*logS_range) to exp(3*logS_range), on a grid equally spread
+        in log scale. The number of grids evaluated is determined by
+        SNR_bins. If you increase logS_range, it is recommended to increase
+        SNR_bins accordingly, otherwise the SNR values evaluated might be
+        too sparse, causing the posterior SNR estimations to be clustered
+        around the bins.
+    SNR_bins: integer. Default: 21
```
lcnature: Done

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2460,1350 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+        One approach is for users to provide time series which they consider
+        as nuisance but exist in the noise (such as head motion).
+        The other way is to take the first n_nureg principal components
+        in the residual after subtracting the response to the design matrix
+        from the data, and use these components as the nuisance regressor.
+        This flag is for the second approach. If turned on,
+        PCA or factor analysis will be applied to the residuals
+        to obtain new nuisance regressors in each round of fitting.
+        These two approaches can be combined. If the users provide nuisance
+        regressors and set this flag as True, then the first few principals
```
lcnature: fixed

brainiak/reprsimil/brsa.py
```diff
@@ -867,38 +1324,258 @@ def _fit_RSA_UV(self, X, Y, X0,
...
+        X = np.concatenate(X)
+        X0 = np.concatenate(X0)
+        return X, X0, total_log_p
+
+    def _score(self, Y, design, beta, scan_onsets, beta0, rho_e, sigma_e,
+               rho_X0, sigma2_X0):
+        """ Given the data Y, and the spatial pattern beta0
+            of nuisance time series, return the cross-validated score
+            of the data Y given all parameters of the subject estimated
+            during the fist step.
```
lcnature: fixed

brainiak/reprsimil/brsa.py
```diff
@@ -798,30 +1235,30 @@ def _fit_RSA_UV(self, X, Y, X0,
...
                 logger.debug(
                     'current GP[1]:{}'.format(current_GP[1]))
                 # We start the length scale for intensity with
-                # a small value. A heuristic is 5 percentile of
+                # a small value. A heuristic is 2 percentile of
                 # all the square differences. But it should not be
                 # smaller than 0.5. This limit is set in case
                 # many voxels have close to equal intensities,
-                # which might render 5 percential to 0.
+                # which might render 2 percential to 0.
```
lcnature: fixed

brainiak/reprsimil/brsa.py
```diff
@@ -464,51 +873,78 @@ def _prepare_data_XY(self, X, Y, D, F):
...
-        X0TX0, X0TDX0, X0TFX0 = self._make_templates(D, F, X0, X0)
-        XTX0, XTDX0, XTFX0 = self._make_templates(D, F, X, X0)
-        X0TY, X0TDY, X0TFY = self._make_templates(D, F, X0, Y)
-
-        return X0TX0, X0TDX0, X0TFX0, XTX0, XTDX0, XTFX0, \
-            X0TY, X0TDY, X0TFY, X0, n_base
+                        'reflet these components.')
+        return X_DC, X_base, idx_DC
 
     def _make_sandwidge(self, XTX, XTDX, XTFX, rho1):
```
lcnature: Done

brainiak/reprsimil/brsa.py
```diff
@@ -464,51 +873,78 @@ def _prepare_data_XY(self, X, Y, D, F):
...
             no_DC means not inserting regressors for DC components
             into nuisance regressor.
-            It will only take effect if X0 is not None.
+            It will only take effect if X_base is not None.
         """
-
-        X_base = scipy.linalg.block_diag(*map(np.ones, run_TRs)).T
-        res = np.linalg.lstsq(X_base, X)
+        X_DC = self._gen_X_DC(run_TRs)
+        res = np.linalg.lstsq(X_DC, X)
```
lcnature: fixed

brainiak/reprsimil/brsa.py
```diff
@@ -30,24 +32,158 @@
...
 
 __all__ = [
-    "BRSA",
+    "BRSA", "GBRSA", "prior_GP_var_inv_gamma", "prior_GP_var_half_cauchy"
 ]
 
 
-class BRSA(BaseEstimator):
+def prior_GP_var_inv_gamma(y_invK_y, n_y, tau_range):
+    """ Imposing a prior onto the variance (tau^2) of the Gaussian
```
lcnature: fixed

brainiak/reprsimil/brsa.py
```diff
@@ -62,22 +198,26 @@ class BRSA(BaseEstimator):
...
 
         \\beta_i \\sim N(0,(s_{i} \\sigma_{i})^2 U)
 
     Parameters
     ----------
-    n_iter : int, default: 200
+    n_iter : int. Default: 50
         Number of maximum iterations to run the algorithm.
-    rank : int, default: None
+    n_iter_inner : int, default: 6
```
lcnature: changed to accepting a dictionary now.

brainiak/reprsimil/brsa.py
```diff
@@ -87,106 +227,220 @@ class BRSA(BaseEstimator):
...
-        This is experimental and slow. I find that when SNR is genrally low,
-        smoothness can be overestimated. But I think this is better than
-        not imposing any regularization.
-    GP_inten: boolean, defualt: False
-        Whether to include a kernel definved over the intensity of image.
+        This is relatively slow for big ROI. We find that when SNR
+        is generally low, smoothness can be overestimated.
+        But such regularization may reduce variance in the estimated
+        SNR map and similarity matrix.
+    GP_inten: boolean. Defualt: False
```
lcnature: added more explanation now

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2460,1350 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+        # X to experiment conditions and other neural activity.
+        # However, in fit(), we keep the tradition of scikit-learn that
+        # X is the input data to fit and y, a reserved name not used, is
+        # the label to map to from X.
+        assert self.SNR_bins >= 5 and self.rho_bins >= 5, \
+            'At least 5 bins are required to perform the numerical'\
+            ' integration over SNR and rho'
+        assert self.logS_range * 6 / self.SNR_bins < 0.5, \
+            'The minimum grid of log(SNR) should not be larger than 0.5.'\
+            ' Please consider increasing SNR_bins or reducing logS_range'
```
lcnature: The first is now in test_gbrsa.py. The second is hard to define for the exponential distribution as a prior. The way I am doing is chunking [0, inf) to finite bins, and choose the expected mean of the exponential distribution of each bin as the grid to integrate. So the first one is by default large than 0.

brainiak/reprsimil/brsa.py
lcnature: Hi @mshvartsman I reverted the default optimizer of GBRSA to BFGS for now. The reason is when I was trying the L-BFGS-B, for some data set the outer loop of fitting does not converge within 80 iterations when using L-BFGS-B while BFGS typically can finish in ~ 40 iterations, which is within the current default n_iter. I think L-BFGS-B might work well if I choose the appropriate n_iter and the 'maxiter' parameter for scipy.optimize.minimize used in the inner loop of fitting. But with the lack of testing what parameters are good for L-BFGS-B to have a good chance of convergence, I leave the default to 'BFGS' for now but keep the docstring to say that L-BFGS-B and CG are possible alternatives.

brainiak/reprsimil/brsa.py
```diff
@@ -30,24 +32,158 @@
...
+            + np.sqrt(n_y**2 * tau_range**4 + (2 * n_y + 8)
+                      * tau_range**2 * y_invK_y + y_invK_y**2))\
+        / 2 / (n_y + 2)
+    log_ptau = scipy.stats.halfcauchy.logpdf(
+        tau2**0.5, scale=tau_range)
+    return tau2, log_ptau
+
+
+def Ncomp_BIC_Minka(X):
+    """ This implements the BIC metric written in Minka 2000:
```
lcnature: function removed.

brainiak/reprsimil/brsa.py
lcnature: an option of whether to do z-scoring for the purpose of calculating nuisance regressor from residual and calculating the number of nuisance regressors is added as an option.

brainiak/reprsimil/brsa.py
```diff
@@ -87,106 +227,220 @@ class BRSA(BaseEstimator):
...
-        optimizer.
-    rand_seed : int, default: 0
+    optimizer: str or callable. Default: 'BFGS'
+        The optimizer to use for minimizing cost function.
+        We use 'BFGS' as a default. Users can try other strings
+        corresponding to optimizer provided by scipy.optimize.minimize,
+        or a custom optimizer.
+    rand_seed : int. Default: 0
         Seed for initializing the random number generator.
+    anneal_speed: float. Default: 20
```
lcnature: This needs further test to decide whether annealing is necessary. I prefer deferring this test for now.

brainiak/reprsimil/brsa.py
lcnature: I decided to still leave it for the user to do z-scoring of the whole data or not before fitting. I personall feel that if training and testing data have slightly different variance in each voxel, z-scoring them separately might cause different scaling among different voxels, then transform() and score() might be impacted, especially transform() which relies on the assumption that the patterns of activation/nuisance activity stays constant across runs.

brainiak/reprsimil/brsa.py
```diff
@@ -421,7 +830,7 @@ def _prepare_DF(self, n_T, scan_onsets=None):
             D = self._D_gen(n_T)
```
lcnature: It is fine to flip the names. But it requires flipping all parameters related to these (XTDX and XTFX, etc.). I think doing this in a separate PR can help make sure that all variables are changed correctly, because this PR already has too many change to the code and may be hard to track these changes.

mihaic: Then please create an issue for this future flipping.

brainiak/reprsimil/brsa.py
lcnature: Since the other method for determining number of components is also based on PCA, the default for nureg_method has changed to PCA now.

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2460,1350 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+    n_nureg: int or string. Default: 'BIC'
+        Number of nuisance regressors to use in order to model signals
+        shared across voxels not captured by the design matrix.
+        This number is in addition to any nuisance regressor that the user
+        has already provided.
+        If set to 'BIC', the number of nuisance regressor will be
+        automatically determined based on BIC metric
+        of probabalistic PCA model. (Minka 2000, NIPS)
+    nureg_method: string, 'PCA', 'FA', 'ICA' or 'SPCA'. Default: 'ICA'
+        The method to estimate the shared component in noise across voxels.
```
lcnature: fixed.

brainiak/reprsimil/brsa.py
```diff
@@ -1659,3 +2460,1350 @@ def _loglike_AR1_singpara(self, param, XTX, XTDX, XTFX, YTY_diag,
...
+        If DC_single is True, only one regressor of constant 1 will be
+        used for the whole dataset. This might be desirable if you
+        believe the average image intensity might not scale with the
+        same proportion for different voxels across scan. In other words,
+        it is possible that some part of the brain is more vulnerable to
+        change in baseline intensity due to facts such as
+        field inhomogeneity. Setting DC_single to True will force the
+        nuisance regressors automatically estimated from residuals to
+        capture this.
+    logS_range: float. Default: 1.0
```
lcnature: It has been made clear to be truncated distribution in the SNR_prior argument now. For the second question, log normal distribution here assumes the mean on log scale is 0 (1 in linear scale). The actual mean does not matter because `L_` will scale accordingly. I think it is reasonable to use symmetric range on both sides on the log scale.

lcnature:
@mshvartsman All comments are addressed except for 3: changing names of matrices D/F would be done in a future PR to make it easy to check that all other variable names depending on these are changed accordingly. Checking the necessity of annealing will be deferred for now. The code will not do z-scoring of data for the user, but option to z-score the residual for the purpose of determining number of nuisance regressors and calculating these regressors is provided. Thanks!

brainiak/reprsimil/brsa.py
lcnature: this is changed to baseline_single instead.

brainiak/reprsimil/brsa.py
```diff
@@ -1226,7 +2090,98 @@ def _fit_diagV_GP(
                 break
 
         return current_vec_U_chlsk_l, current_a1, current_logSNR2,\
-            current_GP, X0
+            current_GP, X_res
+
+    def _fit_null(self, Y, X_base, scan_onsets=None):
```
mshvartsman: For future refactoring it may be possible to just call `fit` here with a design matrix of 1's rather than have separate method (though it will require changing the error checking in `fit` also).

tests/reprsimil/test_gbrsa.py
```diff
@@ -0,0 +1,364 @@
...
+        and np.isclose(np.min(SNR_grids), 0) and np.all(np.diff(SNR_grids) > 0), \
+        'SNR_grids or SNR_weights not correct for log normal prior'
+
+
+    s = brainiak.reprsimil.brsa.GBRSA(SNR_prior='exp')
+    SNR_grids, SNR_weights = s._set_SNR_grids()
+    assert np.all(SNR_grids >= 0) and np.isclose(np.sum(SNR_weights), 1) and np.all(SNR_weights > 0) \
+        and np.all(np.diff(SNR_grids) > 0), \
+        'SNR_grids or SNR_weights not correct for exponential prior'
+
```
mshvartsman: Somewhere around here there need to be tests of the numerical integration (against known closed-form, or against code validated elsewhere like from `scipy.integrate`), and/or there should be math somewhere showing that the numerical integration tricks are correct.

mshvartsman:
With addition of tests of the numerical integration code, plus some changes to the eigenvalue cutoff code @lcnature is doing now, I am happy to have this merged. Sorry it took so long.

tests/reprsimil/test_gbrsa.py
lcnature: Thanks! This is now added as `test_grid_flatten()` in test_gbrsa.py. Unfortunately this is a long test code, because we have to repeat many of the steps done by `_fit_RSA_marginalized` in the test code in order to set up all the tests. Additional check for flattening 2-D grids to 1-D grid is included in this function as well.

brainiak/reprsimil/brsa.py
lcnature: I have not verified that such calling would replace `_fit_null`. But there is no urgency for this.

mihaic:
Thank you, @mshvartsman.

@lcnature, now we only need a blurb for the release notes in a new RST file named "docs/newsfragments/194.feature". It can be as simple as single sentence naming the new method added.

lcnature:
@mihaic Done. Please check it out. Thanks!
Thanks for the careful review @mshvartsman !

brainiak/reprsimil/brsa.py
```diff
@@ -85,113 +272,288 @@ class BRSA(BaseEstimator):
...
-    init_iter: how many initial iterations to fit the model
+    tau_range: float. Default: 5.0
+        The reasonable range of the standard deviation
+        of log(SNR). This range should not be too
+        large. 5 is a loose range.
+        When a Gaussian Process is imposed on the log(SNR),
+        this parameter is used in a half-Cauchy prior
+        on the standard deviation, or an inverse-Gamma prior
+        on the variance of the GP.
+    tau2_prior: callable[[float, int, float], [float, float]],
```
lcnature: Thanks @mihaic I forgot how I wrote this line. It looks like it only takes [float, int, float]. So I will delete the second part of the line as well.

brainiak/reprsimil/brsa.py
```diff
@@ -85,113 +272,288 @@ class BRSA(BaseEstimator):
...
         self.GP_inten = GP_inten
         self.tol = tol
         self.auto_nuisance = auto_nuisance
         self.n_nureg = n_nureg
-        self.verbose = verbose
+        self.nureg_zscore = nureg_zscore
+        if auto_nuisance:
+            assert (type(n_nureg) is str and
+                    n_nureg in ['opt']) \
+                or (type(n_nureg) is int and n_nureg > 0), \
```
lcnature: fixed. But the test for being positive integer is still kept.

brainiak/reprsimil/brsa.py
```diff
@@ -19,8 +19,10 @@
  .. [Cai2016] "A Bayesian method for reducing bias in neural
     representational similarity analysis",
     M.B. Cai, N. Schuck, J. Pillow, Y. Niv,
-    Neural Information Processing Systems (NIPS) 29, 2016.
-    http://papers.nips.cc/paper/6131-a-bayesian-method-for-reducing-bias-in-neural-representational-similarity-analysis
+    Advances in Neural Information Processing Systems 29, 2016, 4952--4960
+    Available at:
+    http://papers.nips.cc/paper/6131-a-bayesian-method-for-reducing-bias-in-neural-representational-similarity-analysis.pdf
+    Some extensions not described in the paper have been made here.
```
lcnature: fixed.

brainiak/reprsimil/brsa.py
```diff
@@ -85,113 +272,288 @@ class BRSA(BaseEstimator):
...
+        both the responses to design matrix and the user-supplied nuisance
+        regressors will be used in addition to the nuisance regressors
+        provided by the users.
         Note that nuisance regressor is not required from user. If it is
-        not provided, DC components for each run will be used as nuisance
-        regressor in the initial fitting.
-    n_nureg: int, default: 6
+        not provided, DC components for each run will be included as nuisance
+        regressor regardless of the auto_nuisance parameter.
+    n_nureg: int or string. Default: 'opt'
```
lcnature: fixed.

brainiak/reprsimil/brsa.py
```diff
@@ -368,7 +354,7 @@ class BRSA(BaseEstimator, TransformerMixin):
         this parameter is used in a half-Cauchy prior
         on the standard deviation, or an inverse-Gamma prior
         on the variance of the GP.
-    tau2_prior: callable[[float, int, float], [float, float]],
+    tau2_prior: Callable[[float, int, float]],
```
mihaic: The return type of the `Callable` is missing:
https://docs.python.org/3/library/typing.html#typing.Callable

brainiak/reprsimil/brsa.py
```diff
@@ -2912,8 +2902,15 @@ class GBRSA(BRSA):
...
-        The normalized pseuso-SNR of all voxels.
-        They are normalized such that the geometric mean is 1
+        The pseuso-SNR of all voxels. If SNR_prior='lognormal',
+        the geometric mean of nSNR\_ would be approximately 1.
+        If SNR_prior='unif', all nSNR\_ would be in the range of (0,1).
+        If SNR_prior='exp' (default), the range of values would vary
+        depending on the data and SNR_bins, but many should have low
+        values with few voxels with high values.
+        Note that this attribute can not be interpreted as true SNR,
+        but the relative ratios between voxel indicates the contribution
```
mihaic: Typo: voxels.

brainiak/utils/utils.py
```diff
@@ -601,3 +601,32 @@ def _read_stimtime_AFNI(stimtime_files, n_C, n_S, scan_onoff):
...
+        parameter as scipy.stats.expon.pdf
+
+    Parameters
+    ----------
+    a: float
+        The starting point of the interval in which the center of mass
+        is calculated for exponential distribution.
+    b: float
+        The ending point of the interval in which the center of mass
+        is calculated for exponential distribution.
```
mihaic: `scale` is missing.

brainiak/reprsimil/brsa.py
```diff
@@ -284,12 +270,12 @@ class BRSA(BaseEstimator, TransformerMixin):
         Note that nuisance regressor is not required from user. If it is
         not provided, DC components for each run will be included as nuisance
         regressor regardless of the auto_nuisance parameter.
-    n_nureg: int or string. Default: 'opt'
+    n_nureg: int or None. Default: None
```
lcnature: Do you mean, writing it as:
`n_nureg: Optional[int]`, period?
Does it also apply to other parameters whose defaults are `None`, such as `rank`, `random_state`, `space_smooth_range` and `inten_smooth_range`?

brainiak/reprsimil/brsa.py
lcnature: Sorry, fixed

brainiak/reprsimil/brsa.py
lcnature: fixed

brainiak/utils/utils.py
lcnature: fixed

brainiak/reprsimil/brsa.py
```diff
@@ -22,7 +22,22 @@
     Advances in Neural Information Processing Systems 29, 2016, 4952--4960
     Available at:
     http://papers.nips.cc/paper/6131-a-bayesian-method-for-reducing-bias-in-neural-representational-similarity-analysis.pdf
-    Some extensions not described in the paper have been made here.
+    Some extensions beyond the paper have been made here.
+    More specifically:
+    (1) spatial noise correlation (or alternatively
+    considered as signals of intrinsic fluctuation not related to tasks);
+    (2) new fitting procedure which marginalize all voxel-specific
```
lcnature: fixed

brainiak/reprsimil/brsa.py
mihaic: I would write:
`n_nureg: int`

This is also what the documentation I linked says ("An optional argument with a default needn’t use the Optional qualifier on its type annotation (although it is inferred if the default is None).") And also what you have already for `inten_smooth_range` and `space_smooth_range`.

mihaic:
Thank you very much for this massive contribution, @lcnature! After keeping you waiting so long for the merge, I hope you find some consolation in knowing you helped us improve the BrainIAK contribution process.

PR #196, opened 2017-03-15
Added shape to searchlight

mjanderson09:


brainiak/searchlight/searchlight.py
```diff
@@ -24,6 +24,35 @@
 ]
 
 
+def Diamond5x5x5():
+    """Diamond3x3
```
yidawang: This one line description will cause confusion because it is inconsistent with the method name. Please clarify.

examples/searchlight/example_searchlight.py
```diff
@@ -90,10 +93,36 @@ def sfn(l, msk, myrad, bcast_var):
...
   for (cnt, img) in enumerate(global_outputs):
     plt.imshow(img,cmap='hot',vmin=0,vmax=1)
-    plt.savefig('img' + str(cnt) + '.png')
+    plt.savefig('cube_img' + str(cnt) + '.png')
     plt.clf()
 
+# Searchlight using diamond-shaped searchlight region
+
+# Create searchlight object
+sl = Searchlight(sl_rad=2, max_blk_edge=5)
```
yidawang: I assume that `Diamond5x5x5` only works for masks with radius >= 2. Is it correct? If yes, please note it somewhere.

mjanderson09:
Will move the masking operation into the searchlight.

PR #197, opened 2017-03-22
Fix bugs before major pull request is approved

lcnature:
Since the pull request of adding GBRSA and score/transform function can take a while, I want to have this minor pull request while the major one is being reviewed.
1. fix the bug in posterior estimation of beta in BRSA.
2. changed the code where scipy.linalg.block_diag is called because its treatment of empty matrix changed last November. This issue also causes the failure of test in the pull request of distributed searchlight.
3. speed up test_brsa.py


mihaic:
Please follow the naming convention for modules, even for example files, and even for notebook files:
https://www.python.org/dev/peps/pep-0008/#package-and-module-names

You could use `bayesian_rsa_example.ipynb`.

lcnature:
Done.

PR #198, opened 2017-03-23
[dev] Fix LDFLAGS for LLVM on MacOS

mihaic:


mihaic:
The lack of "rpath" caused the recent build failure when Brew changed the LLVM version to 4.0.

PR #199, opened 2017-03-23
Dictionary learning

TuKo:
This PR implements the work from INRIA on Dictionary learning for atlas parcellation.
Example is not included as it is planned to be added later when Cameron's paper is ready.

### DIFF
--- a/brainiak/funcalign/dl.py
+++ b/brainiak/funcalign/dl.py
@@ -164,7 +164,10 @@ class MSDL(BaseEstimator, TransformerMixin):
                              .format(len(X)))
 
         if self.kappa < 0.0 or self.kappa >= 1.0:
-            raise ValueError("Kappa should be in range (0,1)")
+            raise ValueError("Kappa should be in range (0,1).")
+
+        if self.lam <= 0.0 or self.mu <= 0.0:
+            raise ValueError("Lambda and Mu regularization parameters should positive.")
 
         # Prepare the laplacian operator for this data
         self.L_ = self._create_laplacian_operator(R)
--- a/brainiak/funcalign/dl.py
+++ b/brainiak/funcalign/dl.py
@@ -1,0 +1,449 @@
+#  Copyright 2016 Intel Corporation
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


mihaic:
@cameronphchen, @hejiaz, you should be familiar with this. Could you please review?

brainiak/funcalign/dl.py
```diff
@@ -0,0 +1,581 @@
...
+The implementation is based on the following publication:
+
+.. [Varoquaux2011] "Multi-subject Dictionary Learning to Segment an Atlas of
+   Brain Spontaneous Activity",
+   G. Varoquaux, A. Gramfort, F. Pedregosa, V. Michel, .B. Thirion
+   22nd International Conference Information Processing in Medical Imaging
+   (IPMI 2011)
+   http://rd.springer.com/chapter/10.1007/978-3-642-22092-0_46
+
+"""
```
mihaic: Please add a paragraph stating this code implements a different algorithm from nilearn.

cameronphchen:
@mihaic @TuKo , sure! I can do the review.

mihaic:
@cameronphchen, please remember this PR.

TuKo:
@cameronphchen 
I checked for convergence of the algorithm with raider in an unusual way.
This method requires the position of the voxels in the brain, compared to SRM that doesn't. 

So unfortunately I don't have data to test for significance results of this method and I'm expecting to add an example further on, when working on the survey paper.

cameronphchen:
@TuKo @mihaic  If that's the case, I don't think we should merge this into the master branch at this point since we can't guarantee its correctness. In other words, we haven't had a success story of using this code. I can work on the example later on, but it's probably better we merge this with the example together. What do you think?

TuKo:
Ok, we will wait for your example to accept the PR.

mihaic:
@cameronphchen, how is the example coming along?

cameronphchen:
@mihaic:  Hejia @hejiaz  has tested the code but there seem to be some issues with the code leading to incorrect results. We have discussed this with Javier @TuKo.

mihaic:
@cameronphchen, @hejiaz, thanks for your help. Javier and I decided to close this PR because the algorithm it implements has been superseded by better algorithms, so it would not be of interest to BrainIAK users. Feel free to reopen if you disagree.

cameronphchen:
Sounds good to me. Thanks!

PR #201, opened 2017-03-29
Substantial edits to how noise is calculated. Some edits to how noise…

CameronTEllis:
… is created. Altered test and example to suit update

### DIFF
--- a/brainiak/utils/fmrisim.py
+++ b/brainiak/utils/fmrisim.py
@@ -57,14 +57,12 @@ Show the brain as it unfolds over time with a given opacity.
 import logging
 
 from itertools import product
-import os
 import nitime.algorithms.autoregressive as ar
 import math
 import numpy as np
 from pkg_resources import resource_stream
 from scipy import stats
 import scipy.ndimage as ndimage
-import nibabel
 
 __all__ = [
     "generate_signal",
@@ -947,7 +945,7 @@ def calc_noise(volume,
 
     # Calculate the fwhm on a subset of volumes
 
-    if volume.shape[3]>100:
+    if volume.shape[3] > 100:
         # Take only 100 shuffled TRs
         trs = np.arange(volume.shape[3])
         np.random.shuffle(trs)
--- a/brainiak/utils/fmrisim.py
+++ b/brainiak/utils/fmrisim.py
@@ -684,15 +684,15 @@ def _calc_fwhm(volume,
                voxel_size=[1.0, 1.0, 1.0],
                ):
     """ Calculate the FWHM of a volume
-    Takes in a volume and mask and outputs the FWHM (mm) of each TR of this
+    Takes in a 3d volume and mask and outputs the FWHM (mm) of this
     volume for the non-masked voxels
 
     Parameters
     ----------
-    volume : multidimensional array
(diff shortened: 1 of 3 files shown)


brainiak/utils/fmrisim.py
```diff
@@ -1168,6 +1230,17 @@ def _generate_noise_spatial(dimensions,
         Generates the spatial noise volume for these parameters
     """
 
+    if len(dimensions) == 4:
+        return
+
+    # Convert from fwhm to sigma (relationship discovered empirical, only an
+    #  approximation up to sigma=0-5 which corresponds to fwhm=0-8, relies
+    # on an assumption of brain size).
+    def logcurve(x, a, b, c):
```
mihaic: Use the standard docstring format, please.

brainiak/utils/fmrisim.py
```diff
@@ -678,90 +681,106 @@ def apply_signal(signal_function,
...
+        # Pull out this coordinate
+        x, y, z = coordinates[i]
+
+        # Is this within the mask?
+        if mask[x, y, z] > 0:
+            # For each xyz dimension calculate the squared
+            # difference of this voxel and the next
+
+            inrange = (x < dimensions[0] - 1)
+            inmask = inrange and (mask[x + 1, y, z] > 0)
```
mihaic: Use `_` to separate words in names.

mihaic:
@cbaldassano, you are listed as an author. Could you please tell us if the changes make sense?

brainiak/utils/fmrisim.py
```diff
@@ -50,7 +50,9 @@
 Show the brain as it unfolds over time with a given opacity.
 
 
- Authors: Cameron Ellis (Princeton) 2016
+ Authors:
+ Cameron Ellis (Princeton) 2016-2017
+ Chris Boldassano (Princeton) 2016-2017
```
cbaldassano: Although I think this is an improved spelling, it is unfortunately "bald" not "bold"

brainiak/utils/fmrisim.py
```diff
@@ -678,90 +681,106 @@ def apply_signal(signal_function,
...
+
+    # Find the variance
+    d_var = np.divide((d_sq - np.divide(np.power(d_sum, 2),
+                                        d_count)), (np.add(d_count, -1)))
+
+    o_var = np.divide(-1, (4 * np.log(1 - (0.5 * d_var / v_var))))
+    fwhm3 = np.sqrt(o_var) * 2 * np.sqrt(2 * np.log(2))
+    fwhm = np.prod(np.multiply(fwhm3, voxel_size)) ** (1 / 3)
 
     return fwhm
```
cbaldassano: This is the function I helped write, looks good to me! There might be ways to make this a little faster (perform a single shift operation rather than per-voxel) but we should only go down that road if it is currently running too slow.

CameronTEllis: To run this on a single TR with 64x64x48 voxels takes 1.002s. In calc_noise I subsample the trs to avoid this taking more than 100 TRs. At the moment this is a sufficient work around.

mihaic:
@CameronTEllis, update to the latest changes in master and we can merge the PR.

PR #209, opened 2017-04-21
Refactor ISFC to use the new I/O and image modules

mihaic:


mihaic:
@cbaldassano, I think the changes to ISFC are minor. Could you please review?

mihaic:
@cbaldassano, thanks for the review.

PR #214, opened 2017-04-25
[htfa] Refactor matrix inversion

mihaic:
Using NumPy implementation, which is not slower.


xiazhu:
looks good to me :)

mihaic:
Thanks for the review, @xiazhu.

PR #215, opened 2017-04-25
[searchlight] Switch from pathos to multiprocess

### LINKED ISSUE #149: Refactor code to remove Pathos

mihaic:
Fixes #149.

### DIFF
--- a/brainiak/searchlight/searchlight.py
+++ b/brainiak/searchlight/searchlight.py
@@ -12,6 +12,7 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+import multiprocess
 import numpy as np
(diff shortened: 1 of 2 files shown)


mihaic:
Abandoning in favor of using `multiprocessing`.

PR #217, opened 2017-04-27
Random permutation of FCMA

yidawang:
1. add random voxel permutation to FCMA;
2. add more debugging and information logs;
3. minor tweak for docstring format, etc.

Thanks in advance for reviewing.

### DIFF
--- a/brainiak/fcma/preprocessing.py
+++ b/brainiak/fcma/preprocessing.py
@@ -22,11 +22,20 @@ import numpy as np
 import logging
 from scipy.stats.mstats import zscore
 from mpi4py import MPI
+from enum import Enum
 
 from ..image import mask_images, multimask_images
 
 
 logger = logging.getLogger(__name__)
+
+__all__ = [
+    "RandomType",
+    "prepare_fcma_data",
+    "generate_epochs_info",
+    "prepare_mvpa_data",
+    "prepare_searchlight_mvpa_data",
+]
 
 
 def _separate_epochs(activity_data, epoch_list):
@@ -37,9 +46,9 @@ def _separate_epochs(activity_data, epoch_list):
 
     Parameters
     ----------
-    activity\_data: list of 2D array in shape [nVoxels, nTRs]
+    activity_data: list of 2D array in shape [nVoxels, nTRs]
         the masked activity data organized in voxel*TR formats of all subjects
-    epoch\_list: list of 3D array in shape [condition, nEpochs, nTRs]
+    epoch_list: list of 3D array in shape [condition, nEpochs, nTRs]
         specification of epochs and conditions
         assuming all subjects have the same number of epochs
         len(epoch_list) equals the number of subjects
@@ -83,8 +92,72 @@ def _separate_epochs(activity_data, epoch_list):
     return raw_data, labels
 
 
+def _randomize_single_subject(data, seed=None):
+    """Randomly permute the voxels of the subject.
+
+     The subject is organized as Voxel x TR,
+     this method shuffles the voxel dimension.
+
+    Parameters
+    ----------
+    data: 2D array in shape [nVxels, nTRs]
+        Activity data.
+    seed: Optional[int]
+        Seed for random state used implicitly for shuffling.
+
+    Returns
+    data: 2D array in shape [nVxels, nTRs]
+        Activity data with the voxel dimension shuffled.
+    """
+    if seed is not None:
+        np.random.seed(seed)
+    np.random.shuffle(data)
+    return data
+
+
+def _randomize_subject_list(data_list, random):
+    """Randomly permute the voxels of a subject list.
+
+     The method shuffles the subject one by one according to
+     the random type. If RandomType.NORANDOM, return the
+     original list.
+
+    Parameters
+    ----------
+    data_list: list of 2D array in shape [nVxels, nTRs]
+        Activity data list.
+    random: RandomType
+        Randomization type.
+
+    Returns
+    data_list: list of 2D array in shape [nVxels, nTRs]
+        (Randomized) activity data list.
+    """
+    if random == RandomType.REPRODUCIBLE:
+        data_list = [_randomize_single_subject(data, seed=idx)
+                     for idx, data in enumerate(data_list)]
+    elif random == RandomType.UNREPRODUCIBLE:
+        data_list = [_randomize_single_subject(data)
+                     for data in data_list]
+    return data_list
+
+
+class RandomType(Enum):
+    """Define the random types as enumeration
+
+    NORANDOM means do not randomize the data;
+    REPRODUCIBLE means randomize the data with a fixed seed so that the
+    permutation holds between different runs;
+    UNREPRODUCIBLE means truly randomize the data which returns different
+    results in different runs.
+    """
+    NORANDOM = 0
+    REPRODUCIBLE = 1
+    UNREPRODUCIBLE = 2
+
+
 def prepare_fcma_data(images, conditions, mask1, mask2=None,
-                      comm=MPI.COMM_WORLD):
+                      random=RandomType.NORANDOM, comm=MPI.COMM_WORLD):
     """Prepare data for correlation-based computation and analysis.
 
     Generate epochs of interests, then broadcast to all workers.
@@ -102,6 +175,9 @@ def prepare_fcma_data(images, conditions, mask1, mask2=None,
         If it is not specified, the method will assign None to the returning
         variable raw_data2 and the self-correlation on raw_data1 will be
         computed
+    random: Optional[RandomType]
+        Randomize the data within subject or not.
+        Default NORANDOM
     comm: MPI.Comm
         MPI communicator to use for MPI operations.
 
@@ -127,9 +203,11 @@ def prepare_fcma_data(images, conditions, mask1, mask2=None,
             activity_data1, activity_data2 = zip(*multimask_images(images,
                                                                    masks,
                                                                    np.float32))
+            _randomize_subject_list(activity_data2, random)
             raw_data2, _ = _separate_epochs(activity_data2, conditions)
         else:
             activity_data1 = list(mask_images(images, mask1, np.float32))
+        _randomize_subject_list(activity_data1, random)
         raw_data1, labels = _separate_epochs(activity_data1, conditions)
         time1 = time.time()
     raw_data_length = len(raw_data1)
@@ -170,7 +248,7 @@ def generate_epochs_info(epoch_list):
 
     Returns
     -------
-    epoch\_info: list of tuple (label, sid, start, end).
+    epoch_info: list of tuple (label, sid, start, end).
         label is the condition labels of the epochs;
         sid is the subject id, corresponding to the index of raw_data;
         start is the start TR of an epoch (inclusive);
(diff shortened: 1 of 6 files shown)


brainiak/fcma/preprocessing.py
```diff
@@ -127,9 +203,11 @@ def prepare_fcma_data(images, conditions, mask1, mask2=None,
             activity_data1, activity_data2 = zip(*multimask_images(images,
                                                                    masks,
                                                                    np.float32))
+            _randomize_subject_list(activity_data2, random)
             raw_data2, _ = _separate_epochs(activity_data2, conditions)
         else:
             activity_data1 = list(mask_images(images, mask1, np.float32))
+        _randomize_subject_list(activity_data1, random)
```
xoltar: _randomize_single_subject modifies data in place (and then returns the same array). _randomize_subject_list returns a new list, containing old arrays that have been modified in place, which is why these lines work even though you're ignoring the return value. I would say either write them both to work in-place, or write them both to return new values. Mixing the two makes for confusion. If you decide you want to have _randomize_single_subject return a new array instead of working in-place, consider numpy.random.permutation instead of shuffle.

brainiak/fcma/preprocessing.py
```diff
@@ -83,8 +92,72 @@ def _separate_epochs(activity_data, epoch_list):
...
 
+def _randomize_single_subject(data, seed=None):
+    """Randomly permute the voxels of the subject.
+
+     The subject is organized as Voxel x TR,
+     this method shuffles the voxel dimension.
+
+    Parameters
+    ----------
+    data: 2D array in shape [nVxels, nTRs]
```
xoltar: nVoxels

brainiak/fcma/preprocessing.py
```diff
@@ -83,8 +92,72 @@ def _separate_epochs(activity_data, epoch_list):
...
+        Activity data.
+    seed: Optional[int]
+        Seed for random state used implicitly for shuffling.
+
+    Returns
+    data: 2D array in shape [nVxels, nTRs]
```
xoltar: nVoxels

brainiak/fcma/preprocessing.py
```diff
@@ -292,8 +374,20 @@ def prepare_searchlight_mvpa_data(images, conditions, data_type=np.float32):
 
     for sid, f in enumerate(images):
         data = f.get_data().astype(data_type)
+        [d1, d2, d3, d4] = data.shape
+        if random == RandomType.REPRODUCIBLE:
+            data = _randomize_single_subject(data.reshape((d1 * d2 * d3, d4)),
```
xoltar: You could make this reshaping logic part of _randomize_single_subject (i.e. have it flatten, randomize, then reshape into whatever shape the original array was).

brainiak/fcma/preprocessing.py
```diff
@@ -303,7 +397,7 @@ def prepare_searchlight_mvpa_data(images, conditions, data_type=np.float32):
                 processed_data[:, :, :, idx] = \
                     np.mean(data[:, :, :, epoch[2]:epoch[3]], axis=3)
 
-        logger.info(
+        logger.debug(
             'file %s is loaded and processed, with data shape %s' %
```
xoltar: logger does formatting itself, you can just do:
`logger.debug('file %s is loaded and processed, with data shape $s', f.get_filename(), data.shape)`

brainiak/fcma/voxelselector.py
```diff
@@ -163,7 +163,8 @@ def _master(self):
             the length of array equals the number of voxels
         """
         logger.info(
-            'Master starts to allocate tasks'
+            'Master at rank %d starts to allocate tasks' %
```
xoltar: Same logging comments apply to this file.

brainiak/io.py
```diff
@@ -59,6 +62,10 @@ def load_images_from_dir(in_dir: Union[str, Path], suffix: str = "nii.gz",
     files = sorted(in_dir.glob("*" + suffix))
     for f in files:
         yield nib.load(str(f))
+        logger.info(
+            'file %s is read' %
```
xoltar: Here too. No need for str(), either.

brainiak/fcma/preprocessing.py
yidawang: done

brainiak/fcma/preprocessing.py
yidawang: done

brainiak/fcma/preprocessing.py
yidawang: Thanks for the suggestion! Somehow I ignored the return value here although I should put it as `activity_data1 = _randomize_subject_list(activity_data1, random)`. I would prefer working in-place, let me modify `_randomize_subject_list` accordingly.

brainiak/fcma/preprocessing.py
yidawang: I didn't do this because I would like to make the input argument `data` of `_randomize_single_subject` an nVoxels x nTRs 2D array for clarity purpose.

brainiak/fcma/preprocessing.py
yidawang: done

brainiak/fcma/voxelselector.py
yidawang: done

brainiak/io.py
yidawang: `f` is a path here, which is not supported to be the argument of `numpy.load()`, and presumably `nibabel.load()`, for numpy version before 1.12. @mihaic should be able to further justify this.

mihaic: Right. However, `Path`s work with string formatting without explicit casting to `str`, e.g.:
```"test %s" % pathlib.Path("/")```

Comment (yidawang) [inline brainiak/io.py]: OK, just realized that Bryn was referring the `str` usage below this line :) Done.

File: brainiak/io.py [hunk 8]
```diff
@@ -59,6 +62,10 @@ def load_images_from_dir(in_dir: Union[str, Path], suffix: str = "nii.gz",
     files = sorted(in_dir.glob("*" + suffix))
     for f in files:
         yield nib.load(str(f))
+        logger.info(
+            'file %s is read',
+            f
+        )
```
mihaic: I would say you should use the debug level for the messages you added in "io.py".

Also, do not break lines if they do not exceed the maximum line length.

Finally, the other message placement, before `yield`, makes more sense, but I would change the wording to "Starting to read [...]" and put a full stop at the end of the message.

examples/fcma/mvpa_voxel_selection.py
```diff
@@ -53,11 +53,24 @@
             'mask size: %d' %
             np.sum(mask)
         )
-        images = io.load_images_from_dir(data_dir, extension)
+        images = io.load_images_from_dir(data_dir, suffix=extension)
```
mihaic: It makes more sense to rename `extension` to `suffix` here.

examples/fcma/voxel_selection.py
```diff
@@ -41,16 +41,30 @@
     extension = sys.argv[2]
     mask_file = sys.argv[3]
     epoch_file = sys.argv[4]
-    images = io.load_images_from_dir(data_dir, extension)
+    images = io.load_images_from_dir(data_dir, suffix=extension)
```
mihaic: Same.

brainiak/fcma/preprocessing.py
```diff
@@ -266,10 +345,13 @@ def prepare_searchlight_mvpa_data(images, conditions, data_type=np.float32):
         Condition specification.
     data_type
         Type to cast image to.
+    random: Optional[RandomType]
+        Randomize the data within subject or not.
+        Default NORANDOM
```
yidawang: done

brainiak/io.py
```diff
@@ -25,12 +25,15 @@
 
 import nibabel as nib
 import numpy as np
+import logging
```
yidawang: done

examples/fcma/voxel_selection.py
yidawang: done

examples/fcma/mvpa_voxel_selection.py
yidawang: done

brainiak/io.py
yidawang: done, but I would put some brief `logger info` to indicate in the beginning of some stages (reading file, apply masking, separating epochs, etc.)

mihaic: Sure, but not in "io.py".

yidawang: Why not? I think within the io methods are the best places to go.
BTW, I am feeling that `io.load_images_from_dir` and `io.load_images` should be merged.

yidawang: anyway, I will leave `io.py` intact

mihaic: Because they are not useful. Why would a function log an info-level message saying it has been called every time it is called? Such messages could be emitted by the caller, if relevant.

We can discuss merging the load functions, but I would rather not, because "_dir" requires a suffix as well, so we would need to explain that.

yidawang: well, the suffix can be an optional argument. Let's discuss this in the next standup.

brainiak/io.py
```diff
@@ -58,6 +61,9 @@ def load_images_from_dir(in_dir: Union[str, Path], suffix: str = "nii.gz",
         in_dir = Path(in_dir)
     files = sorted(in_dir.glob("*" + suffix))
     for f in files:
+        logger.debug(
+            'starts to read file %s', f
+        )
```
mihaic: Please write sentences in log messages, at least in this file.

yidawang: done

brainiak/fcma/preprocessing.py
```diff
@@ -83,8 +92,72 @@ def _separate_epochs(activity_data, epoch_list):
...
+    Returns
+    data_list: list of 2D array in shape [nVxels, nTRs]
+        (Randomized) activity data list.
+    """
+    if random == RandomType.REPRODUCIBLE:
+        for i in range(len(data_list)):
+            data_list[i] = _randomize_single_subject(data_list[i], seed=i)
+    elif random == RandomType.UNREPRODUCIBLE:
+        for i in range(len(data_list)):
+            data_list[i] = _randomize_single_subject(data_list[i])
```
xoltar: This is better, but still mixed usage. If you're going to do the change in place, don't return a value. If you're going to return a value, it should be a new value. So if you want to use in-place, it should look like:
```
for subject in data_list:
    _randomize_single_subject(subject)
```
So neither _randomize_single_subject nor this function should return a value. Also, note since you're modifying in place, you don't need to assign to indexes, so you can use the simpler for loop.

yidawang: done

brainiak/fcma/preprocessing.py
```diff
@@ -83,8 +92,71 @@ def _separate_epochs(activity_data, epoch_list):
...
+     this method shuffles the voxel dimension.
+
+    Parameters
+    ----------
+    data: 2D array in shape [nVoxels, nTRs]
+        Activity data.
+    seed: Optional[int]
+        Seed for random state used implicitly for shuffling.
+
+    Returns
```
xoltar: Good! Now just need to remove the Returns section from docs of both of these functions since they don't return anything anymore.

PR #218, opened 2017-04-27
Revert "[dev] Switch from `virtualenv` to `venv`"

mihaic:
Reverts IntelPNI/brainiak#213

Unfortunately, `venv` is broken in EPEL, which is used for Python 3 by most RHEL and CentOS users, including ourselves on our own clusters:
https://bugzilla.redhat.com/show_bug.cgi?id=1319963
https://bugzilla.redhat.com/show_bug.cgi?id=1263057

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ matrix:
       before_install: *macos_before_install
 
 install:
-    - python3 -m pip install -U pip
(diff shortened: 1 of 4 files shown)


mihaic:
Actually, @mjanderson09 discovered the problem.

PR #220, opened 2017-04-28
Distributed srm

mjanderson09:


brainiak/funcalign/srm.py
```diff
@@ -132,6 +144,8 @@ class SRM(BaseEstimator, TransformerMixin):
     rho2_ : array, shape=[subjects]
         The estimated noise variance :math:`\\rho_i^2` for each subject
 
+    comm_ : mpi4py.MPI.Intracomm
+        The MPI communicator containing the data
```
mihaic: `comm` should be used instead of `comm_` everywhere, because it is given as a parameter, not computed from data:
http://scikit-learn.org/stable/developers/contributing.html#instantiation
http://scikit-learn.org/stable/developers/contributing.html#estimated-attributes

brainiak/funcalign/srm.py
```diff
@@ -258,11 +291,18 @@ def _init_structures(self, data, subjects):
         rho2 = np.zeros(subjects)
 
         trace_xtx = np.zeros(subjects)
+
         for subject in range(subjects):
-            mu.append(np.mean(data[subject], 1))
+            np.random.seed(subject)
```
mihaic: How does this interact with `self.rand_seed`?

tests/funcalign/test_srm.py
```diff
@@ -191,3 +223,4 @@ def test_det_srm():
     print("Test: different number of samples per subject")
 
 
+test_can_instantiate()
```
mihaic: Wrap this call in `if __name__ == "__main__"`, please.

brainiak/funcalign/srm.py
mihaic: Not addressed.

TuKo: Check issue #114 
Is it right to initialize the distributed SRM randomly?

mjanderson09: I removed this random seed per subject code. Instead, each SRM class instantiation can set a random seed (as before). This means for multiple ranks, each rank can potentially set a different seed. I don't think we can guarantee exactly deterministic results for the distributed implementation due to differences in parallel scheduling per run.

mihaic:
Jenkins, retest this please.

mihaic:
Jenkins, retest this please.

TuKo:
@mjanderson09 great job!
Please, consider adding the TS-SVD and TS-QR code as well. We should allow for reproducibility of the IEEE BigData algorithm. Also, all SRM flavors will benefit from it, and some other methods as well.

Don't need to do it now. You can do it in a different PR.

Thank you!

tests/funcalign/test_srm.py
```diff
@@ -62,6 +63,9 @@ def test_can_instantiate():
 
     # Check that runs with 2 subject
     s.fit(X)
+    sr_v0_4 = np.load('sr_v0_4.npy')
+    diff_v0_4 = np.linalg.norm(sr_v0_4 - s.s_)
+    assert(diff_v0_4 < 1e-5)
```
mihaic: Shouldn't the results be element-wise equal, i.e., `np.allclose(sr_v0_4, s.s_)`, maybe with an `atol` value?

tests/funcalign/test_srm.py
```diff
@@ -62,6 +63,9 @@ def test_can_instantiate():
 
     # Check that runs with 2 subject
     s.fit(X)
+    sr_v0_4 = np.load('sr_v0_4.npy')
```
mihaic: Add this to imports:
`from pathlib import Path`

Then:
`Path(__file__).parent / "sr_v0_4.npy"`

mihaic:
@TuKo, what say you?

TuKo:
@TuKo says NIPS, sorry...

tests/funcalign/test_srm.py
```diff
@@ -190,4 +194,4 @@ def test_det_srm():
         model.fit(X)
     print("Test: different number of samples per subject")
 
-
+test_can_instantiate()
```
mihaic: Please guard this call with `if __name__ == "__main__"`.

TuKo:
:+1:

mihaic:
For the record, here is the effect of the changes on the accuracy in the image prediction example:
```
v0.4:
SRM: The average accuracy among all subjects is 0.657143 +/- 0.058576
Det. SRM: The average accuracy among all subjects is 0.657143 +/- 0.070891

master (i.e., per-subject random number generators):
SRM: The average accuracy among all subjects is 0.664286 +/- 0.076097
Det. SRM: The average accuracy among all subjects is 0.673214 +/- 0.068721

distributed_srm:
SRM: The average accuracy among all subjects is 0.664286 +/- 0.076097
Det. SRM: The average accuracy among all subjects is 0.673214 +/- 0.068721
```

PR #221, opened 2017-04-28
Searchlight improvement

mjanderson09:


brainiak/searchlight/searchlight.py
```diff
@@ -24,6 +25,50 @@
...
+
+        """
+        self.rad = rad
+        self.data = np.ones((2*rad+1, 2*rad+1, 2*rad+1), dtype=np.bool)
+
+
+class Diamond:
+    """Diamond
+
+    Searchlight shape which is a (2*rad+1)^3 diamond
```
mihaic: Does this mean the diamond contains exactly that number of voxels?

brainiak/searchlight/searchlight.py
```diff
@@ -46,8 +91,13 @@ def __init__(self, sl_rad=1, max_blk_edge=10):
 
         """
         self.sl_rad = sl_rad
-        self.max_blk_edge = max_blk_edge
+        if max_blk_edge == -1:
```
mihaic: Why not let the user specify `max_blk_edge=sl_rad*3` as parameter?

tests/searchlight/test_shapes.py
```diff
@@ -0,0 +1,80 @@
+#  Copyright 2016 Intel Corporation
```
mihaic: I think theses tests should be part of "test_searchlight.py".

tests/searchlight/test_shapes.py
```diff
@@ -0,0 +1,80 @@
...
+#  limitations under the License.
+
+import numpy as np
+from mpi4py import MPI
+import sys
+
+from brainiak.searchlight.searchlight import Searchlight
+from brainiak.searchlight.searchlight import Diamond
+
+def test_cube():
```
mihaic: These tests should be called `test_searchlight_with_cube/diamond`, because they test the shapes' usage in the searchlight, not the shape classes themselves.

brainiak/searchlight/searchlight.py
```diff
@@ -24,6 +25,73 @@
...
+        """Constructor
+
+        Parameters
+        ----------
+
+        rad: radius, in voxels, of the sphere inscribed in the
+             searchlight cube, not counting the center voxel
+
+        """
+        super(Cube, self).__init__(rad)
```
mihaic: You should call `super()` with no arguments (also in `Diamond`):
https://docs.python.org/3/library/functions.html#super

brainiak/searchlight/searchlight.py
```diff
@@ -24,6 +25,73 @@
...
+        self.rad = rad
+        self.data = np.ones((2*rad+1, 2*rad+1, 2*rad+1), dtype=np.bool)
```
mihaic: Could you please document `data` as a class attribute in the superclass?

tests/searchlight/test_shapes.py
mihaic: Not addressed.

mihaic:
@yidawang, it looks like the size of the output array has changed for 26 to 23 in "test_mvpa_voxel_selection". Any idea why?

tests/searchlight/test_shapes.py
mjanderson09: OK I've made this change. Thank you.

brainiak/searchlight/searchlight.py
```diff
@@ -345,18 +425,22 @@ def _singlenode_searchlight(l, msk, mysl_rad, bcast_var):
                                                           mysl_rad:-mysl_rad,
                                                           mysl_rad:-mysl_rad]
 
+            def check_mask(mask_cube):
```
mihaic: Please make this a method and document it.

brainiak/searchlight/searchlight.py
```diff
@@ -24,6 +25,80 @@
...
+class Shape:
+    """Shape
+
+    Searchlight shape which is contained in a (2*rad+1)^3 cube
+
+    Attributes
+    ----------
+
+    data_ : a 3D boolean numpy array of size (2*rad+1)^3 which is set
+            to True within the boundaries of the desired shape
```
mihaic: As discussed, this is a mask, so it should be named appropriately, e.g., `mask_`.

mihaic: Could you please change the array description to use shape, like elsewhere in BrainIAK, e.g., `of shape (2*rad+1, 2*rad+1, 2*rad+1)`?

brainiak/searchlight/searchlight.py
```diff
@@ -345,18 +425,22 @@ def _singlenode_searchlight(l, msk, mysl_rad, bcast_var):
                                                           mysl_rad:-mysl_rad,
                                                           mysl_rad:-mysl_rad]
 
+            def check_mask(mask_cube):
+                res = mask_cube[self.shape]
+                return np.any(res) and np.all(res)
```
mihaic: Isn't `np.all` sufficient?

mjanderson09: No, because it returns true for an empty array.

mihaic: Then how about:
```python
return len(res) > 0 and np.all(res)
```

PR #225, opened 2017-05-12
dev: Remove Pathos and multiprocess dependencies

mihaic:
Use multiprocessing instead. Adjust searchlight and fcma to use
pickle-able functions.


mihaic:
Jenkins, retest this please.

Issue #228, opened 2017-06-05
Document usage of examples


mihaic:
How are users supposed to get the examples? Via git? Should we provide a link to an archived version?

mihaic:
Fixed in #231.

PR #231, opened 2017-06-07
docs: Document usage of examples

### LINKED ISSUE #93: Add link to Jupyter notebook example in README
This will immediately show what can be done with our package.

mihaic:
Fixes #93, #228.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -101,6 +101,20 @@ as long as you see ``Successfully installed`` at the end.
 .. _issue #61:
    https://github.com/IntelPNI/brainiak/issues/61
 
+
+Examples
+========
+
+We include BrainIAK usage examples in the examples directory of the code
+repository, e.g., `funcalign/srm_image_prediction_example.ipynb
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -101,6 +101,20 @@ as long as you see ``Successfully installed`` at the end.
 .. _issue #61:
    https://github.com/IntelPNI/brainiak/issues/61
 
+
+Examples
+========
+
+We include BrainIAK usage examples in the examples directory of the code
```
yidawang: We should point out that this is an example of SRM in BrainIAK. For other components currently we also have example codes available but not in Notebook format.

mihaic: I think the name shown there is self explanatory: "srm_image_prediction_example". And we do have notebooks for other examples as well.

yidawang: then how about providing more example links?

mihaic: I would rather keep the README short. This is only an invitation for users to look in the examples directory.

yidawang: OK, sounds good

PR #232, opened 2017-06-14
docs: Document requirements.txt for examples

mihaic:
Also add some missing requirements for some examples.

### DIFF
--- a/examples/searchlight/requirements.txt
+++ b/examples/searchlight/requirements.txt
@@ -1,3 +1,2 @@
 matplotlib
 nilearn
-notebook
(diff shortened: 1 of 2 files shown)


examples/funcalign/requirements.txt
```diff
@@ -1,2 +1,3 @@
 matplotlib
+nilearn
```
yidawang: Could you explain why this dependency just appeared, or why it was not here before?

examples/searchlight/requirements.txt
```diff
@@ -0,0 +1,2 @@
+matplotlib
+nilearn
```
yidawang: As above.

examples/funcalign/requirements.txt
mihaic: It should have been there already (i.e., this PR fixes a bug).

examples/searchlight/requirements.txt
mihaic: Same.

PR #239, opened 2017-07-13
changed behavior of rand_seed in BRSA

lcnature:
Per request by @mshvartsman, rand_seed argument now defaults to None, and BRSA won't reset random seed if None is provided.
The suppression of warning is also removed.


lcnature:
Hi @mihaic Sorry I am a bit confused. How should numpy.random.RandomState replace numpy.random.random()? I thought the former sets the state while the latter just return a random number.
Thanks!

lcnature:
Does that mean I should call this `random_state = check_random_state(random_state)`, and from then on, replace all `np.random` with `random_state`?

mihaic:
`RandomState` objects also have several methods for generating random numbers. `RandomState.random_sample()` is a good replacement for `numpy.random.random()`:
https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.RandomState.random_sample.html

You do not need to use `check_random_state`.

lcnature:
Please see if the new commit is correct. Thanks!

brainiak/reprsimil/brsa.py
```diff
@@ -686,7 +688,7 @@ def _fit_RSA_UV(self, X, Y, X0,
         n_C = np.size(X, axis=1)
         l_idx = np.tril_indices(n_C)
 
-        np.random.seed(self.rand_seed)
+        self.random_state_ = check_random_state(self.random_state)
```
mihaic: Please document this attribute in the class docstring. Could be something like this:
```
random_state_: `RandomState`
    Random number generator initialized using random_state
```

lcnature: Thanks. Just added exactly this text.

PR #250, opened 2017-07-29
dev: Remove MANIFEST.in

### LINKED ISSUE #90: MANIFEST.in is not sufficient to include files in installation
https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files

mihaic:
The only file we list, LICENSE, is included in the sdist regardless of
MANIFEST.in, thanks to setuptools_scm. Furthermore, bdist_wheel ignores
MANIFEST.in:
https://packaging.python.org/tutorials/distributing-packages/#manifest-in

Fixes #90.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,0 @@
-include LICENSE


TuKo:
:+1:

PR #258, opened 2017-08-11
dev: Run flake8 on test code

mihaic:
Because we allow `print` in test code, I had to explicitly list the code
we want flake8 to ignore. I think this is better in the long run anyway.

I changed most of the tests to comply. However, I did add a few `noqa`
markers, mostly for complexity. I will make separate issues for each
package for removing them.

Partially addresses #19.

### DIFF
--- a/run-checks.sh
+++ b/run-checks.sh
@@ -18,8 +18,9 @@ set -e
 set -o pipefail
 
 flake8 --config setup.cfg brainiak
+flake8 --config tests/.flake8 tests
 mypy --ignore-missing-imports brainiak tests/[!_]*
-rst-lint *.rst | { grep -v "is clean.$" || true; }
+rst-lint ./*.rst | { grep -v "is clean.$" || true; }
 towncrier --version=100 --draft > /dev/null 2>&1 \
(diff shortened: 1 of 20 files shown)


setup.cfg
```diff
@@ -1,5 +1,16 @@
...
+    # Defaults
+    # https://gitlab.com/pycqa/flake8/blob/master/src/flake8/defaults.py
+    E121,
+    E123,
+    E126,
+    E226,
+    E24,
+    E704,
+    W503,
+    W504,
```
yidawang: Why didn't we have these ignores in the first place?

mihaic: We did have them; they were there by default.

yidawang: Oh, I thought you just put them into this `setup.cfg` file.

mihaic: `flake8` has a list of codes it ignores by default, visible at the link I put in `setup.cfg`. Because we did not explicitly list any codes, we were implicitly using the default list.

Now that we want to ignore the print function error (`T003`) we need to explicitly declare the ignored codes, at least for the test code. I thought it would be better to have the list explicit for the main code as well, so they don't diverge.

PR #260, opened 2017-08-14
make n_subj required

xiazhu:
small change to make n_subj a required parameter.

### DIFF
--- a/brainiak/factoranalysis/htfa.py
+++ b/brainiak/factoranalysis/htfa.py
@@ -67,7 +67,7 @@ class HTFA(TFA):
     K : int
         Number of factors to compute.
 
-    n_subj : int, default: 2
+    n_subj : int
         Total number of subjects in dataset.
 
     max_global_iter : int, default: 10
--- a/brainiak/factoranalysis/htfa.py
+++ b/brainiak/factoranalysis/htfa.py
@@ -64,12 +64,11 @@ class HTFA(TFA):
     Parameters
     ----------
 
-    R : list of 2D arrays, element i has shape=[n_voxel, n_dim]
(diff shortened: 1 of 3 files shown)


brainiak/factoranalysis/htfa.py
```diff
@@ -64,23 +64,19 @@ class HTFA(TFA):
     Parameters
     ----------
 
-    R : list of 2D arrays, element i has shape=[n_voxel, n_dim]
```
mihaic: Was `R` not used?

brainiak/factoranalysis/htfa.py
```diff
@@ -64,23 +64,19 @@ class HTFA(TFA):
...
-        Each element in the list contains the scanner coordinate matrix
-        of fMRI data of one subject.
-
-    K : int, default: 50
+    K : int
         Number of factors to compute.
 
+    n_subj : int, default: 2
```
mihaic: We should remove the default from the docstring too.

As an aside, having to change the docstring now is a good reason to not put the default in the docstring in the first place, given that it is visible in the signature and Sphinx adds it to the HTML docs.

brainiak/factoranalysis/htfa.py
xiazhu: R is not required for construct HTFA. It is a parameter to fit() function.

brainiak/factoranalysis/htfa.py
xiazhu: updated with new commit.

mihaic:
Thank you for the quick PR, @xiazhu.

As another aside, the test giving the last error you were seeing is a good example of a too specialized test.

PR #263, opened 2017-08-22
Fix test_gbrsa.py and test_brsa.py

lcnature:
This is the update after the noqa comments were inserted to the test files.


mihaic:
It looks good, but before merging, Mingbo will look into the reason for the existing tolerance being exceeded.

lcnature:
Hopefully this fixed the problem. I think the reason of the previous mismatch between the grid integration using `scipy.stats.lognorm` and my own implementation might be due to a bias when I set the grids of SNR's. So now I changed the implementation to use `scipy.stats.lognorm`'s functions to set all these grids.

PR #264, opened 2017-08-24
small fixes after Hackathon at UT Austin

yidawang:
1. change process_num default value to be more conservative; 
2. add notes for possible Python multiprocessing error (happened on Lonestar5 of TACC)
3. fix a bug in the Python version of correlation normalization

### DIFF
--- a/brainiak/fcma/voxelselector.py
+++ b/brainiak/fcma/voxelselector.py
@@ -94,11 +94,16 @@ class VoxelSelector:
         the number of hardware threads,
         i.e. the number returned by cpu_count().
         If 0, cross validation will not use python multiprocessing.
-        NOTE: on some compute cluster, python multiprocessing launched
-        via MPI on a remote machine may result in child processes disfunct.
 
     master_rank: int, default 0
         The process which serves as the master
+
+    Note
+    ----
+
+        On some compute cluster (e.g. lonestar5 at TACC),
+        python multiprocessing launched
+        via MPI on a remote machine may result in child processes defunct.
     """
     def __init__(self,
                  labels,
@@ -118,7 +123,7 @@ class VoxelSelector:
         self.num_voxels2 = raw_data2[0].shape[1] \
             if raw_data2 is not None else self.num_voxels
         self.voxel_unit = voxel_unit
-        self.process_num = process_num
+        self.process_num = np.min((process_num, multiprocessing.cpu_count()))
         if self.process_num == 0:
             self.use_multiprocessing = False
         else:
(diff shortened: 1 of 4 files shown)


brainiak/fcma/voxelselector.py
```diff
@@ -91,7 +91,11 @@ class VoxelSelector:
     process_num: Optional[int]
         The number of processes used in cross validation.
         If None, the number of processes will equal
-        the number of hardware threads.
+        the number of hardware threads,
+        i.e. the number returned by cpu_count().
+        If 0, cross validation will not use python multiprocessing.
+        NOTE: on some compute cluster, python multiprocessing launched
+        via MPI on a remote machine may result in child processes disfunct.
```
mihaic: I am not sure what you mean by "disfunct". Can you please rephrase this? Most importantly, please link to a description of the issue, perhaps even create a new issue if it is not yet clear what is happening.

Finally, please make this note using the `Note` RST syntax, so that it stands out in the HTML docs, like in SRM:
http://pythonhosted.org/brainiak/brainiak.funcalign.html#brainiak.funcalign.srm.SRM.random_state_
https://github.com/IntelPNI/brainiak/blob/master/brainiak/funcalign/srm.py#L81

brainiak/fcma/voxelselector.py
```diff
@@ -103,7 +107,7 @@ def __init__(self,
                  raw_data,
                  raw_data2=None,
                  voxel_unit=64,
-                 process_num=None,
+                 process_num=4,
```
mihaic: This should be limited by the number of CPUs, i.e., max(4, cpus).

tests/fcma/test_voxel_selection.py
```diff
@@ -46,14 +46,14 @@ def test_voxel_selection():
...
-                               [-0.91005552, -0.64101928, 1.48848987,
-                                -0.78406328, -0.83182675]]]
+        expected_fake_corr = [[[1.06988919, 0.51641309, -0.46790636,
+                                -1.31926763, 0.2270218],
+                               [-1.22142744, -1.39881694, -1.2979387,
+                                1.05702305, -0.6525566],
+                               [0.89795232, 1.27406132, 0.36460185,
+                                0.87538344, 1.5227468],
+                               [-0.74641371, -0.39165771, 1.40124381,
+                                -0.61313909, -1.0972116]]]
```
mihaic: Why do we need to change this values?

brainiak/fcma/voxelselector.py
yidawang: Defunct... Sorry for the typo

I cannot find a reference to this issue, but will follow up with the TACC guys. The `Note` syntax is fixed.

brainiak/fcma/voxelselector.py
yidawang: I think it should be min(4, cpus). Fixed.

tests/fcma/test_voxel_selection.py
yidawang: I have double checked with the results generated from an independent process.

PR #266, opened 2017-09-19
Update to eventseg

cbaldassano:
Minor update to eventseg to support additional use case, added ipython notebook eventseg example, also updated download link in isfc example


brainiak/eventseg/event.py
```diff
@@ -345,7 +364,11 @@ def find_events(self, testing_data, var=None, scramble=False):
         """
 
         if var is None:
-            var = self.event_var_
+            if self.event_var_ is None:
```
mihaic: The test should check the existence of the attribute, not whether it is `None`. The documentation is not perfectly clear, but the code is:
> the existence of parameters with trailing _ is used to check if the estimator has been fitted.
http://scikit-learn.org/stable/developers/contributing.html#parameters-and-init
```python
    if not all_or_any([hasattr(estimator, attr) for attr in attributes]):
        raise NotFittedError(msg % {'name': type(estimator).__name__})
```
https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/validation.py#L793

mihaic:
You should also update the paper reference to point to Neuron.

PR #270, opened 2017-10-05
WIP: docs: Use sphinx-gallery

### LINKED ISSUE #268: docs: Generate notebook outputs for web display
Currently, we do not show notebook outputs. It would be helpful to do so. However, we do not want to store the outputs in Git because they pollute diffs.

We should investigate the solution used by Scikit-learn:
* No output in Git:
https://github.com/scikit-learn/scikit-learn/blob/master/examples/datasets/plot_digits_last_image.py
https://github.com/scikit-learn/scikit-learn.github.io/blob/master/dev/_downloads/plot_digits_last_image.ipynb
* Output on the web:
http://scikit-learn.org/stable/auto_examples/datasets/plot_digits_last_image.html

mihaic:
This PR only introduces the infrastructure and an eventseg example.
The rest of the examples will need to be converted in future PRs.

Note that the README files have to have the "txt" extension and the
RST titles within have to be formatted hierarchically at least for the
first two levels (`examples` and its subdirectories) so they are
displayed properly in the resulting HTML.

Fixes #268.


mihaic:
@cbaldassano, what do you think about the Sphinx-Gallery idea of creating notebooks from Python files?
https://sphinx-gallery.readthedocs.io/en/latest/tutorials/plot_notebook.html#sphx-glr-tutorials-plot-notebook-py

cbaldassano:
This looks nice, and it would be great to have the example plots in with the documentation.

It would be nice workflow-wise if this was able to take notebooks as input - or have a system that converts notebooks to python files with their syntax, and then passes them to sphinx-gallery. Is that possible?

mihaic:
Good point about the workflow. I'll look for such a converter.

mihaic:
@cbaldassano, it looks like a converter is not coming to Sphinx-Gallery anytime soon for reasons explained on their issue tracker:
https://github.com/sphinx-gallery/sphinx-gallery/issues/245

However, the following script (linked in the issue above) should give reasonable output in my brief experience:
https://gist.github.com/chsasank/7218ca16f8d022e02a9c0deb94a310fe

I can add a link to the script above to the contributors guide. What do you think?

cbaldassano:
Yep that looks good to me. The issues they discuss (incompatibilities with Markdown and magic functions) shouldn't be a big problem.

Looking at the eventseg/HiddenMarkovModels.ipynb notebook, some potential problems could be:
-The "%matplotlib inline" command, which would just need to get dropped
-The wget commands for pulling the example data
-Some of the cells take a little while to run (~10 minutes for the whole notebook)

The biggest of these is the reliance on external data. Should we only use sphinx-gallery for self-contained examples?

mihaic:
I think we can include external data as long as we do the downloading in Python. This can be as simple as `urllib.request.urlretrieve`.

However, run time can be a problem. As we integrate more examples, I think we should create a new make target for building the examples, which should not be called by default. This will take care of another problem I foresee: dependency bloat for development requirements. Right now, each example directory has its own requirements file, but if we build the examples by default, we need to centralize the requirements.

mihaic:
Closing this PR. We still issue #268 to continue the conversation. I think we should use a solution that works directly with Jupyter notebooks, without converters, such as nbsphinx:
https://github.com/spatialaudio/nbsphinx

PR #271, opened 2017-10-14
Added HTFA tutorial notebook and readme

jeremymanning:
I've added an example Jupyter notebook with examples showing how to:
- Fit TFA and HTFA to a sample dataset
- Create some plots of static and dynamic networks

The readme file also includes a link to a Docker image with all of the necessary toolboxes.

Note: I prepared these materials as part of the [MIND Summer School at Dartmouth](https://github.com/Summer-MIND/mind_2017)


buildbot-princeton:

buildbot-princeton:

mihaic:
Jenkins, add to whitelist.

mihaic:
Thanks for the PR, @jeremymanning. How long does it take on your computer to run all the cells? Mine has been stuck on `tfa.get_centers` for more than an hour.

jeremymanning:
I've only run it from the docker image that I posted; it takes about 20 minutes for the full thing to run on my laptop.  It seems like something may be stuck...

mihaic:
I successfully ran the notebook using the MIND Docker image. However, given that it is supposed to be a BrainIAK example, I think it should also run with up-to-date versions of dependencies, most importantly HyperTools and BrainIAK.

Testing in the Docker container with the latest version of HyperTools, I get a deprecation warning (`UserWarning: Model and model_params arguments will be deprecated.`) and then an exception (`MemoryError`).

@jeremymanning, could you please test the notebook with the latest release of HyperTools?
```
pip install --upgrade --no-deps hypertools
pip install hypertools
```

jeremymanning:
Sure, doing this now...

jeremymanning:
Closing-- I'll submit another request with an updated notebook.

jeremymanning:
Oh...actually, maybe I can just re-open this one?

mihaic:
Sure, you can push updates here.

jeremymanning:
Ok-- the master branch on ContextLab/brainiak is updated, so it looks like those will go through here as well...?

mihaic:
Indeed. I'll try the new version right away.

jeremymanning:
Hi @mihaic,

Thanks for reviewing the code.  I have a few comments on your requests:

> Remove the .md file. The container is not required; I ran the notebook outside the MIND container (also, we are adding a BrainIAK-specific Docker image, see #272). Instead, add the Python dependencies that are not BrainIAK dependencies to examples/factoranalysis/requirements.txt. Add a note about FFmpeg being required for the animations (apt/brew install ffmpeg).

The BrainIAK toolbox is already very difficult to install, and the extra step of installing ffmpeg will further reduce the likelihood of anyone using the tutorial.  The Docker image I linked to makes installation easy by providing a container where the necessary libraries are already installed, and that can be loaded onto any of the major platforms.  I am glad to hear you're adding your own Docker image.  I'd be happy to do any of the following:
1. Close this pull request and continue host this tutorial separately from BrainIAK (currently it's part of the [MIND summer school materials](https://github.com/Summer-MIND/mind_2017))
2. Update the readme to use the BrainIAK-specific Docker image, once you have that set up and assuming the tutorial can run in that container.
3. If the BrainIAK docker image doesn't have hypertools, seaborn, nilearn, and/or nibabel installed, I can continue pointing to the MIND Docker image via the readme file.

However, I am not comfortable posting this notebook with no readme or easy way to run it.  I think that will lead to more confusion.  (Let me know how you'd like to proceed; any of these options are OK with me!)

> Make the path for the downloaded data relative, e.g.:
>
> destination = 'pieman_data.zip'
> std_destination = 'standard_brain.nii.gz'

To clarify, you want the data to download to whatever the user's current working directory is, rather than `/mnt/`?  I can do this, but I think the `/mnt/` solution is cleaner-- then the user knows where the downloaded data live, and it'll be the same for everyone running the tutorial.  (This is also why I think the docker image is an important part of the tutorial; it keeps the setup consistent across users.)

> Use ISFC implementation in BrainIAK:
> https://github.com/IntelPNI/brainiak/blob/master/examples/isfc/isfc.py

The BrainIAK implementation of ISFC works on brain images.  In the tutorial ISFC is applied to numpy arrays.  Also, the implementation in the tutorial computes ISFC dynamics, whereas (to my skim of the code) the BrainIAK implementation doesn't.  So as far as I can tell, the two functions are not interchangeable.

> Remove the MPI code, it does not do anything in notebooks. If you want, point users to the HTFA > MPI example:
> https://github.com/IntelPNI/brainiak/blob/master/examples/factoranalysis/run_mpi_htfa.sh

Sure.  My reason for including the MPI code is that I thought users might want to cut and paste the code from the tutorial into their own applications.  (I found the run_mpi_htfa.sh example confusing.)  What if I add a note explaining this?

> Add a comment justifying the necessity of ignoring warnings or remove the warnings filter.

I disagree with this decision-- I think outputting additional warnings makes the code messy, and the warnings in this case are uninformative.

> Finally, it would be nice to avoid re-downloading/unarchiving the data when repeating the execution.

Good idea.

mihaic:
@jeremymanning, thanks for addressing my comments. I would definitely like get your tutorial in BrainIAK.

Regarding Docker, the BrainIAK image has all the dependencies of all examples installed. I will add the dependencies of your example.

Regarding file system paths, it is generally not a good idea to hard code absolute paths in source code. We cannot assume users have a particular file system layout. At the same time, I think users can easily understand paths relative to the example notebook.

Regarding ISFC, I propose you rename the function to make it clear it differs from the BrainIAK implementation and add a comment about the differences. @cbaldassano, your input is appreciated.

Regarding MPI, if you insist, you can leave the code as is. However, you should change the text to make it clear that there is no parallel execution. I would still prefer removing the code and pointing to the MPI example. Also, I would appreciate an issue or PR for improving the MPI example if you find it confusing.

Regarding the warnings, you can leave the filter in, as I was saying initially. However, you should only ignore specific warnings (using `filterwarnings`; a blank filter is like a catch-all `except`) and you should explain why it is there. Something like: "Because of <something about the data>, <NumPy> will generate <some type of warnings>. We can safely ignore them."

cbaldassano:
Jeremy's ISFC function allows for sliding windows, which the brainiak version currently does not support. So renaming it to clarify that this is a sliding-window version should limit the confusion.

jeremymanning:
I can take care of this next week...

mihaic:
@jeremymanning, new year, new commits?

jeremymanning:
I'll get to this soon!

jeremymanning:
@mihaic could you add hypertools to the brainiak docker image?  It's needed for the hypertools tutorial

jeremymanning:
I've updated the notebook as requested.  Key changes:
- removed .md file
- do not download (or generate) files if they already exist
- suppress matplotlib warning with explanation
- removed MPI code
- renamed `ISFC` to `dynamic_ISFC` to highlight difference from main brainiak function
- removed hypertools examples that were not directly related to HTFA
- general cleanup

I've also added several dependencies to the dockerfile and requirements:
- hypertools (used to plot voxel locations in HTFA tutorial)
- python3-tk (used to display figures in jupyter notebooks)
- ffmpeg (used for creating animations)

requirements-dev.txt
```diff
@@ -10,3 +10,4 @@ setuptools_scm
 sphinx
 sphinx_rtd_theme
 towncrier
+hypertools
```
mihaic: Example requirements are not development requirements, therefore this change is not necessary.

setup.py
```diff
@@ -133,6 +133,7 @@ def finalize_options(self):
         'psutil',
         'nibabel',
         'typing',
+        'hypertools'
```
mihaic: Example requirements are not install requirements either, so this is also not necessary.

jeremymanning:
Hi @mihaic,

I've made the requested changes to requirements-dev.txt and setup.py.  I've also updated the HTFA tutorial notebook to use relative (rather than absolute) paths.  The sample data now gets downloaded to the current working directory.

mihaic:
Thank you for your contribution, @jeremymanning!

PR #280, opened 2017-10-25
dev: Blacklist buggy SciPy release

mihaic:


mihaic:
Jenkins, retest this please.

mihaic:
Not sure. The bug will always be there in SciPy 1.0.0, which the user can have installed for some reason.

danielsuo:
Ah fair enough. Maybe we add a note to some section in the main README
about install troubleshooting? Mostly because people don't really read
error messages...(or documentation for that matter...)

mihaic:
The installation always succeed because Pip enforces `scipy!=1.0.0` when installing BrainIAK. If afterward the user changes the environment in incompatible ways, I think a note in the README will not help.

danielsuo:
Oh really? Even if the user has some other version of the package? Did not
know that.

PR #283, opened 2017-10-29
Fix utils.center_mass_exp and added a test code

lcnature:


brainiak/utils/utils.py
```diff
@@ -627,6 +631,9 @@ def center_mass_exp(a, b, scale=1.0):
         distribution.
     """
     assert a < b, 'a must be smaller than b'
+    assert b > 0, 'b must be larger than 0'
+    assert scale > 0, 'scale must be positive'
+    assert a >= 0, 'a cannot be smaller than 0'
```
mihaic: These restrictions should be documented in the parameters section of the function docstring.

mihaic:
@lcnature, @QihongL, will you implement Javier's suggestion? If not, I will merge this PR as is.

QihongL:
I will try that today! Seems to be a good one to get familiarized with the contribution process!

brainiak/utils/utils.py
QihongL: By the way, regarding this section: if a >= 0, and b > a, then b must  > 0, so we don't need to check b > 0. That said, I actually think the current setting return more informative error messages, so I actually like it. Is this what we want?

mihaic: Let's skip the unnecessary check.

mihaic:
Superseded by PR #288. Thanks for the prompt replies, @lcnature.

PR #284, opened 2017-10-30
dev: Fix pr-check for sdist

mihaic:
Currently, pr-check fails when calling run-tests in an sdist directory
because of the editable install detection.

Note that editable install detection is there to minimize testing errors
during development, i.e., the developer testing the checked out code
while having a different version installed.


mihaic:
Jenkins, retest this please.

PR #287, opened 2017-11-02
genre_searchlight distributed

mjanderson09:
Changed genre_searchlight_example.py to run distributed with MPI. I also added a random seed so that I could verify the solution. The verification I did was to look at img75 output for running 1 MPI process vs. 8 MPI processes. The images are attached for both. It appears to be the same. 

img75

img75


mihaic:
Thank you, @mjanderson09.

PR #290, opened 2017-11-06
eventseg: Return float in masked_log

### LINKED ISSUE #289: dtype in EventSeg Cython should be float
Code [here](https://github.com/brainiak/brainiak/blob/master/brainiak/eventseg/_utils.pyx) will fail on the following example

```python
masked_log(np.array([0, 1, 2]))
```

We get the following error:

```
y[i] = float('-inf')
OverflowError: cannot convert float infinity to integer
```

mihaic:
Fixes #289.


mjanderson09:
Impressive!

Issue #291, opened 2017-11-06
Searchlight: Default pool_size=None might not be a good idea for HPC systems


davidt0x:
Hi,

I have been trying to track down reasons for large memory requirements of the searchlight code on research computing clusters at Princeton. I have done the testing on della but this would probably be a common problem on many HPC systems. It seems the default value for the argument pool_size in run_searchlight function is None. This will cause the worker threads to utilize all cores on a machine by default. On a personal machine this may be desirable but on a shared HPC resource this may not be a good idea since a user might not be provisioned all the cores on the node. When trying to run the searchlight example on della this was causing large memory requirements because the default behavior was to try and spawn a process for each core on the machine. In addition, if the code was run under srun with default arguments (a typical method for spawning MPI code under slurm schedulers), a process is spawned per rank, who in turn tries to spawn a processes for each core on the machine. This was the root cause of the large memory errors on della I believe.

I suggest a default value of 1 for pool_size or maybe clearer documentation if the default behavior is desirable on desktop\laptops. For a longer term solution, it may be preferable to avoid using multiprocessing at all and use MPI for the same node parallelization. If I am reading the code correctly then map is being used on each node to distribute work to processes. However, this is causing the blocks to be pickled and sent to each process since they are passed as arguments and not setup in shared memory. This seems basically equivalent to an MPI process. If it was rewritten using MPI this way it would have the added benefit of avoiding process fork which can cause trouble on certain HPC setups that use OpenMPI implementations. Della is one such case it seems but I am sure there are others. 

Dave Turner

mihaic:
@davidt0x, thanks for the effort to get BrainIAK going on Princeton clusters.

The current implementation gives good performance on laptops without extra configuration and without MPI. At the same time, it does not penalize cluster performance, if the properly configured. Having said that, I think there is room for discussion, so I am inviting people to share their thoughts.

Wildcarde:
Would it be possible for the system to test the value of `cat /proc/$$/status | grep Cpus_allowed_list` or similar at startup (if it exists) to see if they have any restrictions on cpu count?

mihaic:
@Wildcarde, what do you think about PR #293?

Wildcarde:
Seems like it should cover it based on the docs.

davidt0x:
I looked at this patch and how it would work on della. It doesn't seem like it will work because apparently os.sched_getaffinity is not available on all systems. Della is apparently one of those systems.

```
>>> import os
[...]

PR #293, opened 2017-11-07
Restrict multiprocessing to available CPUs

mihaic:
See issue #291.


mihaic:
Good question, @mjanderson09. I see no reason for performance to change. In practice, I see the same the same performance on `genre_searchlight_example` with `pool_size=None`:
```
# before
Searchlight Done:  4.976254463195801
# after
Searchlight Done:  4.268961191177368
```

PR #296, opened 2017-11-10
utils: Add usable_cpu_count

mihaic:
We can remove this and rely only on `os.sched_getaffinity` when it
supports the same OSes as `psutil`. See issue #291.

See also https://bugs.python.org/issue26692.

### DIFF
--- a/brainiak/fcma/voxelselector.py
+++ b/brainiak/fcma/voxelselector.py
@@ -106,7 +106,7 @@ class VoxelSelector:
                  raw_data,
                  raw_data2=None,
                  voxel_unit=64,
-                 process_num=None,
+                 process_num=4,
                  master_rank=0):
         self.labels = labels
         self.epochs_per_subj = epochs_per_subj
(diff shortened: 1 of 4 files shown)


brainiak/fcma/voxelselector.py
```diff
@@ -106,7 +106,7 @@ def __init__(self,
                  raw_data,
                  raw_data2=None,
                  voxel_unit=64,
-                 process_num=4,
+                 process_num=None,
```
yidawang: The `process_num` was intentionally set to be 4, a small number, instead of the "usable cpus" in the system because on a cluster node, the "usable cpus" could be over 30, which will likely blow up the memory for having the process duplicate so many copies.

mihaic: The memory for forked processes is allocated copy-on-write. Unless `_cross_validation_for_one_voxel` allocates a lot of data, it should be fine to have lots of processes in the pool.

yidawang: I understand this. Previously we observed memory blowout when `process_num` was set to be "usable cpus". So I would say yes...

xiazhu:
looks good.

davidt0x:
Just wanted to give a heads up that I gave it a test run on della and it seems resolve the issue.

PR #302, opened 2017-11-23
Robust Shared Response Model

TuKo:
The code introduces the Robust SRM, tests and an example with synthetic data.

### DIFF
--- a/brainiak/funcalign/rsrm.py
+++ b/brainiak/funcalign/rsrm.py
@@ -200,10 +200,46 @@ class RSRM(BaseEstimator, ClassifierMixin, TransformerMixin):
         """
         S = np.zeros_like(X)
         R = None
-        for iter in range(self.n_iter):
+        for i in range(self.n_iter):
             R = self.w_[subject].T.dot(X - S)
             S = self._shrink(X - self.w_[subject].dot(R), self.lam)
         return R, S
+
+    def transform_subject(self, X, y=None):
+        """Transform a new subject using the existing model
+
+        Parameters
+        ----------
+
+        X : 2D array, shape=[voxels, samples]
+            The fMRI data of the new subject.
+
+        y : not used
+
+        Returns
+        -------
+
+        w : 2D array, shape=[voxels, features]
+            Orthogonal mapping `W_{new}` for new subject 
+
+        s : 2D array, shape=[voxels, samples]
+            Individual term `S_{new}` for new subject
+        """
+        # Check if the model exist
+        if hasattr(self, 'w_') is False:
+            raise NotFittedError("The model fit has not been run yet.")
+
+        # Check the number of TRs in the subject
+        if X.shape[1] != self.r_.shape[1]:
+            raise ValueError("The number of samples(TRs) does not match the one"
+                             " in the model.")
+
+        s = np.zeros_like(X)
+        for i in range(self.n_iter):
+            w = self._update_transform_subject(X, s, self.r_)
+            s = self._shrink(X - w.dot(self.r_), self.lam)
+
+        return w, s
 
     def _rsrm(self, X):
         """Block-Coordinate Descent algorithm for fitting RSRM.
@@ -242,7 +278,7 @@ class RSRM(BaseEstimator, ClassifierMixin, TransformerMixin):
             logger.info('Objective function %f' % objective)
 
         # Main loop
-        for iter in range(self.n_iter):
+        for i in range(self.n_iter):
             W = self._update_transforms(X, S, R, subjs)
             S = self._update_outliers(X, W, R, subjs, self.lam)
             R = self._update_shared_response(X, S, W, subjs, features, TRs)
--- a/brainiak/funcalign/rsrm.py
+++ b/brainiak/funcalign/rsrm.py
@@ -52,8 +52,8 @@ class RSRM(BaseEstimator, TransformerMixin):
 
     The model is estimated solving the following optimization problem:
 
-    .. math:: \\min_{W_i, S_i, R}\\sum_i \\frac{1}{2}\|X_i - W_i R - S_i\|_F^2 
-    .. math:: + \\gamma\|S_i\|_1 
+    .. math:: \\min_{W_i, S_i, R}\\sum_i \\frac{1}{2}\|X_i - W_i R - S_i\|_F^2
+    .. math:: + \\gamma\|S_i\|_1
     .. math:: s.t. \\qquad W_i^TW_i = I \\quad \\forall i=1 \\dots N
 
     The solution to this problem is obtained by applying a Block-Coordinate
--- a/brainiak/funcalign/rsrm.py
+++ b/brainiak/funcalign/rsrm.py
@@ -35,12 +35,13 @@ __all__ = [
 
 logger = logging.getLogger(__name__)
 
+
 class RSRM(BaseEstimator, ClassifierMixin, TransformerMixin):
     """Robust Shared Response Model (RSRM)
 
     Given multi-subject data, factorize it as a shared response R among all
-    subjects, an orthogonal transform W per subject, and an outlying (individual) 
-    sparse component S per subject:
+    subjects, an orthogonal transform W per subject, and an outlying 
+    (individual) sparse component S per subject:
 
     .. math:: X_i \\approx W_i R + S_i, \\forall i=1 \\dots N
 
@@ -54,7 +55,8 @@ class RSRM(BaseEstimator, ClassifierMixin, TransformerMixin):
         Number of features to compute.
 
     gamma : float, default: 1.0
-        Regularization parameter for the sparseness of the individual components.
+        Regularization parameter for the sparseness of the individual 
+        components.
 
     rand_seed : int, default: 0
         Seed for initializing the random number generator.
@@ -62,7 +64,7 @@ class RSRM(BaseEstimator, ClassifierMixin, TransformerMixin):
 
     Attributes
     ----------
-
+    
     w_ : list of array, element i has shape=[voxels_i, features]
         The orthogonal transforms (mappings) for each subject.
 
@@ -77,7 +79,7 @@ class RSRM(BaseEstimator, ClassifierMixin, TransformerMixin):
 
     Note
     ----
-
+    
         The number of voxels may be different between subjects. However, the
         number of samples for the alignment data must be the same across
         subjects. 
--- a/brainiak/funcalign/rsrm.py
+++ b/brainiak/funcalign/rsrm.py
@@ -54,7 +54,7 @@ class RSRM(BaseEstimator, TransformerMixin):
 
     .. math:: \\min_{W_i, S_i, R}\\sum_i 1/2\|X_i - W_i R - S_i\|_F^2
     .. math:: + \\gamma\|S_i\|_1
-    .. math:: s.t. \\qquad W_i^TW_i = I  \\forall i=1 \\dots N
+    .. math:: s.t. \\qquad W_i^TW_i = I \\quad \\forall i=1 \\dots N
 
     The solution to this problem is obtained by applying a Block-Coordinate
     Descent procedure. More details can be found in [Turek2017]_.
--- a/brainiak/funcalign/rsrm.py
+++ b/brainiak/funcalign/rsrm.py
@@ -52,8 +52,7 @@ class RSRM(BaseEstimator, TransformerMixin):
 
     The model is estimated solving the following optimization problem:
 
-    .. math:: \\min_{W_i, S_i, R}\\sum_i \\frac{1}{2}\|X_i - W_i R - S_i\|_F^2
-    .. math:: + \\gamma\|S_i\|_1
+    .. math:: \\min_{W_i, S_i, R}\\sum_i \\frac{1}{2}\|X_i - W_i R - S_i\|_F^2 + \\gamma\|S_i\|_1 # noqa
     .. math:: s.t. \\qquad W_i^TW_i = I \\quad \\forall i=1 \\dots N
 
     The solution to this problem is obtained by applying a Block-Coordinate
--- a/brainiak/funcalign/rsrm.py
+++ b/brainiak/funcalign/rsrm.py
@@ -488,6 +488,7 @@ class RSRM(BaseEstimator, ClassifierMixin, TransformerMixin):
         """
         A = Xi.dot(R.T)
         A -= Si.dot(R.T)
+        # Solve the Procrustes problem
         U, _, V = np.linalg.svd(A, full_matrices=False)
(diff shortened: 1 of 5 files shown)


TuKo:
@CameronTEllis and @hejiaz, please do you mind reviewing this PR?

Thank you and happy thanksgiving!

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+        Returns
+        -------
+
+        R : array, shape=[features, samples]
+            The updated shared response.
+        """
+        R = np.zeros((features, TRs))
+        for i in range(subjs):
+            R += W[i].T.dot(X[i])
+            R -= W[i].T.dot(S[i])
```
CameronTEllis: I feel like this is such a critical step in RSRM that it should be explained somewhere (that you are subtracting from the shared response a component fitted to each participant)

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+        W : list of array, element i has shape=[voxels_i, features]
+            The orthogonal transforms (mappings) :math:`W_i` for each subject.
+
+        R : array, shape=[features, samples]
+            The shared response.
+
+        S : list of array, element i has shape=[voxels_i, samples]
+            The individual component :math:`S_i` for each subject.
+
+        subjs : int
```
CameronTEllis: Will these ever be something other than len(X)

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+            components.
+
+        Returns
+        -------
+
+        func : float
+            The RSRM objective function evaluated on the parameters to this
+            function.
+        """
+        func = .0
```
CameronTEllis: Some more description of this objective function would be appreciated. For instance, it is testing the difference between the individual data and the fitted response. However, I am unsure why you subtract the S rather than add or why you take the absolute sum of S or why you half the total

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+        return R
+
+    @staticmethod
+    def _update_transform_subject(X, S, R):
+        """Updates the mappings `W_i` for one subject.
+
+        Parameters
+        ----------
+
+        X : array, shape=[voxels, samples]
```
CameronTEllis: Maybe use a different letter, or lower case since this is no longer a list like it is in other functions. same with S below

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+
+        Returns
+        -------
+
+        W : array, shape=[voxels, features]
+            The orthogonal transform (mapping) :math:`W_i` for the subject.
+        """
+        A = X.dot(R.T)
+        A -= S.dot(R.T)
+        U, _, V = np.linalg.svd(A, full_matrices=False)
```
CameronTEllis: This is probably obvious but I am unsure why you are doing the SVD. A comment would help

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
```diff
@@ -0,0 +1,191 @@
...
+    "noise_level = noise_level / (10 ** (snr / 10))\n",
+    "noise_level = np.sqrt(noise_level  / subjects / voxels / samples)\n",
+    "\n",
+    "for s in range(subjects):\n",
+    "    n = noise_level * np.random.randn(voxels, samples)\n",
+    "    data[s] += n\n",
+    "    \n",
+    "S = [None] * subjects\n",
+    "for s in range(subjects):\n",
+    "    S[s] = (np.random.rand(data[s].shape[0], samples) < p) * ((np.random.rand(data[s].shape[0], samples) * amplitude - amplitude/2) )\n",
```
CameronTEllis: Why use uniform noise? I would have thought gaussian noise would be more reasonable?

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
```diff
@@ -0,0 +1,191 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "voxels = 100\n",
+    "samples = 200\n",
+    "subjects = 10\n",
+    "features = 3\n",
+    "snr = 20 # in dB\n",
+    "amplitude = 8.0 # approximate amplitude used for the values in Si\n",
+    "p = 0.05 # probability of being non-zero in Si"
```
CameronTEllis: How does this relate to the gamma value used later? Why are they not the same?

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
```diff
@@ -0,0 +1,191 @@
...
+    "\n",
+    "# Compute noise_sigma from desired SNR\n",
+    "noise_level = noise_level / (10 ** (snr / 10))\n",
+    "noise_level = np.sqrt(noise_level  / subjects / voxels / samples)\n",
+    "\n",
+    "for s in range(subjects):\n",
+    "    n = noise_level * np.random.randn(voxels, samples)\n",
+    "    data[s] += n\n",
+    "    \n",
+    "S = [None] * subjects\n",
```
CameronTEllis: Might be good to make another example where there is no S variability and show that RSRM does as well as SRM

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
TuKo: S is not the noise but the sparse individual component. If we use a Normal distribution, S will mix with the shared signal because sums of Normal distributions are also Normal distributions. In that case you will have a difficult time to identify what comes from where (Shared vs individual). This is also a common assumption for the RPCA.

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+
+    n_iter : int, default: 10
+        Number of iterations to run the algorithm.
+
+    features : int, default: 50
+        Number of features to compute.
+
+    gamma : float, default: 1.0
+        Regularization parameter for the sparseness of the individual
+        components.
```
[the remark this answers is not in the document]
TuKo: done

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+    ----------
+
+    w_ : list of array, element i has shape=[voxels_i, features]
+        The orthogonal transforms (mappings) for each subject.
+
+    r_ : array, shape=[features, samples]
+        The shared response.
+
+    s_ : list of array, element i has shape=[voxels_i, samples]
+        The individual components with outlying data for each subject.
```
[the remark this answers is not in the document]
TuKo: Done

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+
+    random_state_: `RandomState`
+        Random number generator initialized using rand_seed
+
+    Note
+    ----
+
+        The number of voxels may be different between subjects. However, the
+        number of samples for the alignment data must be the same across
```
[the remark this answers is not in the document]
TuKo: Done (everywhere)

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+    def fit(self, X, y=None):
+        """Compute the Robust Shared Response Model
+
+        Parameters
+        ----------
+
+        X : list of 2D arrays, element i has shape=[voxels_i, samples]
+            Each element in the list contains the fMRI data of one subject.
+
+        y : not used
```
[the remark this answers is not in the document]
TuKo: Removed.

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+                "There are not enough samples to train the model with "
+                "{0:d} features.".format(self.features))
+
+        # Check if all subjects have same number of TRs for alignment
+        number_trs = X[0].shape[1]
+        number_subjects = len(X)
+        for subject in range(number_subjects):
+            assert_all_finite(X[subject])
+            if X[subject].shape[1] != number_trs:
+                raise ValueError("Different number of alignment samples "
```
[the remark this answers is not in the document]
TuKo: Done.

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+            Not thread safe.
+        """
+
+        # Init the Random seed generator
+        np.random.seed(self.rand_seed)
+
+        # Draw a random W for each subject
+        W = [random_state.random_sample((voxels[i], features))
+             for i in range(subjs)]
+        for i in range(subjs):
```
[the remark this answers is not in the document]
TuKo: Done.

brainiak/funcalign/rsrm.py
TuKo: Done.

brainiak/funcalign/rsrm.py
TuKo: This is the objective function that defines the problem (eq 2 in paper, will add it to description of method). In fact, Si is adding to the shared response term (Wi*R),  but it is subtracting from the original data Xi. The idea is to make sure that there is data fidelity (Wi*R+Si is approx. Xi). In other words, we try to minimize the distance between Xi and (Wi*R+Si). 
We are not taking the absolute value of Si but we are computing the l1-norm, which is known to obtain sparse solutions. In other words, we are requiring each Si to be sparse.

All the description is in the paper and this is an auxiliary internal function (for logging).

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+            Each element in the list contains the fMRI data for alignment of
+            one subject.
+
+        S : list of array, element i has shape=[voxels_i, samples]
+            The individual component :math:`S_i` for each subject.
+
+        W : list of array, element i has shape=[voxels_i, features]
+            The orthogonal transforms (mappings) :math:`W_i` for each subject.
+
+        subjs : int
```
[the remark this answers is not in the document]
TuKo: Removed

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+            The number of subjects.
+
+        features : int
+            The number of features in the model.
+
+        TRs : int
```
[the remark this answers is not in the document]
TuKo: Removed (BTW is X[0].shape[1])

brainiak/funcalign/rsrm.py
TuKo: Done

brainiak/funcalign/rsrm.py
TuKo: Done

brainiak/funcalign/rsrm.py
TuKo: Done

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
TuKo: p is the probability of an element to be non-zero in the *original* Si (drawn from the model).
gamma is selected to approximately have the same number of non-zeros in the *estimated* Si.
However, there is no reason for them to have the same values. Ones is the probability and the other is a balancing value between regularization terms.

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
TuKo: The idea of the example is to show how to use it. I chose the synthetic example as it is easy to replicate and visually show the improvements. If you set gamma to "infinity", you will be solving exactly the (deterministic) SRM.

brainiak/funcalign/rsrm.py
```diff
@@ -0,0 +1,560 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Robust Shared Response Model (RSRM)
+
+The implementation is based on the following publications:
+
```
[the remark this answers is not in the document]
TuKo: I added an extended description of the method. However, I don't think that the relation to RPCA is relevant to the code/implementation. The idea is that people can understand the code and relate to the paper if there is something unclear.

TuKo:
@CameronTEllis do you have any suggestion or comment for the newest version?

brainiak/funcalign/rsrm.py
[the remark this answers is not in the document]
CameronTEllis: Fair enough

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
CameronTEllis: I might still be misunderstanding but shouldn't the non-zero elements of the original Si be thought of as the individual component that RSRM is aiming to extract? Why would there be a different number in the input vs the output?

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
CameronTEllis: Good

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
CameronTEllis: Fine

brainiak/funcalign/rsrm.py
CameronTEllis: Fair enough

examples/funcalign/rsrm_synthetic_reconstruction.ipynb
TuKo: The math is not translated 1:1. Remember that this is probability, and when you draw a signal there are not exactly 5% of non-zeros. There could be anywhere near the 0.05. Added to that this will be differently with all subjects. So every subject will have approximately that, but not exactly. 
Therefore, even if you can obtain a 1:1 translation from p (range 0 to 1) to gamma (range 0 to +inf), still the signal itself will not necessarily behave in that way. 

Also, this is a particular case of drawing Si's, and you can change it by other distributions as well. In such case, you will have parameters to generate the signals and still gamma to recover it. Dividing between gamma and p, it is a good idea.

CameronTEllis: Okay fair enough I understand

TuKo:
@CameronTEllis do you think there are more changes to be done? If we are done, then don't forget to approve the review. Otherwise let me know.
Thanks!

PR #305, opened 2017-11-27
Solves the failing test of SS-SRM

### LINKED ISSUE #304: Theano-MKL changes break SS-SRM tests due to error
Theano with MKL 2018 requires to define the environment variable MKL_THREADING_LAYER=GNU to run. 
Therefore the ss-srm test is failing.

TuKo:
This PR resolves an issue between Theano and MKL 2018 that requires an environment variable for MKL to run Theano.

Fixes #304 .

### DIFF
--- a/docs/newsfragments/304.trivial
+++ b/docs/newsfragments/304.trivial
@@ -1,0 +1,1 @@
+Solved an issue that test won't pass because of a missing evironment variable for Theano with MKL versions >=2018.
--- a/pr-check.sh
+++ b/pr-check.sh
@@ -125,7 +125,7 @@ python3 -m pip install $ignore_installed -U -r requirements-dev.txt || \
 
 # build documentation
 cd docs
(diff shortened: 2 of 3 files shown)


mihaic:
Jenkins, retest this please.

mihaic:
Jenkins, retest this please.

mihaic:
I could successfully run `pr-check` from my account on the Jenkins machine once. The second time I tried, it failed with the same error seen in Jenkins. Very strange.

TuKo:
Maybe we should move to non-conda? Although that is valuable, there could be some issue with conda packages at the moment.

mihaic:
As a temporary fix, we could switch to Python 3.4 from EPEL, but I think it is important to test on Anaconda as well, given that it is popular with our users. If I cannot fix the issue until the end of the day, I will make the switch.

TuKo:
I agree with you, but macOS is installing with conda, right?
There should be something wrong with the linux version.
Also, we should start fixing the versions of some tools.

PR #308, opened 2017-12-06
dev: Reinstall setuptools in Conda env

mihaic:
Addresses errors such as the one in PR #305.

If the PyPI version of setuptools is newer than the one in Conda and
backwards incompatible, the build process can break.

When used with the `-I` flag, `pip install` will overwrite the Conda
`pkgs` cache, which is used for creating new environments.


mihaic:
This will only pass the tests together with the changes in PR #305. Closing and moving the change there.

PR #309, opened 2017-12-11
Update to deal with issues for convolution and drift (ready for review)

CameronTEllis:
Two open issues reflect problems in fmrisim. One problem was with the convolution incorrectly downsampling — it now takes the mean and also increases the speed of analysis. Another problem was with the drift: it was a sine wave and that didn't make much sense. It now defaults to a combination of cosines.


mihaic:
@CameronTEllis, are you in touch with anyone for a review?

CameronTEllis:
Given that they are small changes, are you able to review them? Or should I ask someone more familiar with the code?

mihaic:
Please ask someone more familiar.

CameronTEllis:
@lcnature Would it be possible for you to find time to review this please?

mihaic:
@lcnature, we would really appreciate your help reviewing this PR.

brainiak/utils/fmrisim.py
```diff
@@ -469,6 +469,16 @@ def generate_stimfunction(onsets,
         # Pull out the onsets, weights and durations, set as a float
         for line in text:
             onset, duration, weight = line.strip().split()
+
+            # Check if the onset is more precise than the temporal resolution
+            dp = len(onset[onset.find('.') + 1:])
```
lcnature: There may be issue if some digits are written in scientific format, e.g., 1.0234e+03. We may want to comment in the docstring that such formatting can cause problem. Or, maybe we can solve it by converting the string to float and find the resolution in floating number.

brainiak/utils/fmrisim.py
```diff
@@ -771,7 +781,19 @@ def convolve_hrf(stimfunction,
...
                  ):
-    """ Convolve the specified hrf with the timecourse
+    """ Convolve the specified hrf with the timecourse. In order
+    to accelerate the (slow) convolution, stimfunction is downsampled (mean
+    filtering) by the specified temporal resolution (e.g. a stimfunction of
+    100000 elements becomes 100 elements if the temporal_resolution is
+    1000). If temporal_resolution is 1 then the output will be the same
+    length as stimfunction.
+    Be aware that if scaling is on and events are very short
+    (> temporal_resolution * tr_duration) then the hrf may or may not come
```
lcnature: What does it mean to be ` (> temporal_resolution * tr_duration)` and "very short"? It seems that for tr_duration=2 and temporal_resolution=1000, this means something is larger than 2000. I suppose this is not the event duration, is it?

brainiak/utils/fmrisim.py
```diff
@@ -771,7 +781,19 @@ def convolve_hrf(stimfunction,
...
+    out as anticipated. This is because very short events would evoke a small
+    absolute response after convolution  but if there are only short events
+    and you scale then this will look identical to a convolution with longer
+    events. In general scaling is useful, which is why it is the default,
+    but be aware of this edge case
```
lcnature: Sorry I forget about this. Can user avoid this issue by setting scaling to false?

brainiak/utils/fmrisim.py
```diff
@@ -804,10 +826,13 @@ def convolve_hrf(stimfunction,
...
 
     """
+    # How will stimfunction be resized
+    stride = int(temporal_resolution * tr_duration)
+    duration = int(stimfunction.shape[0] / stride)
 
     # Generate the hrf to use in the convolution
     if hrf_type == 'double_gamma':
-        hrf = _double_gamma_hrf(temporal_resolution=temporal_resolution)
+        hrf = _double_gamma_hrf(temporal_resolution=1 / tr_duration)
```
lcnature: Although down-sampling stim_function helps speed up the convolution, I worry that this defeats the original purpose of building a smooth double_gamma_hrf. Consider a case of TR=2s, a stimulus of condition A starts at 1.5s and lasts for 0.5s, and a stimulus of condition B starts at 2s and lasts for 0.5s. I think the original implementation would make the design matrix of the two conditions appear as down-sampled version of two HRF with onsets 0.5 sec apart (they would have slightly different shapes, but very close to each other). But the new implementation might make them exactly the same shape, but 2 sec apart. Is this true? If so, maybe the old implementation is more desirable. You may set the default temporal resolution to some lower number such as 100 or 50 to achieve some speed-up with the old implementation. (most often the displays in scanners have refresh rate of just 60 Hz)

brainiak/utils/fmrisim.py
```diff
@@ -1420,6 +1448,11 @@ def _generate_noise_temporal_drift(trs,
     tr_duration : float
         How long in seconds is each volume acqusition
 
+    basis : str
+        What is the basis function for the drift. Could be made of discrete
+        cosines (number of bases scale with duration) or a sine wave with
```
lcnature: You probably meant `tr_duration` by `duration`? And I did not quite understand `with the temporal order`

brainiak/utils/fmrisim.py
CameronTEllis: Wow astute observation! This was a great comment and I have hopefully fixed it now

brainiak/utils/fmrisim.py
CameronTEllis: You are right, this wasn't clear. Hopefully it is more obvious now. The point is that if the events last less than a TR you may get weird results

brainiak/utils/fmrisim.py
CameronTEllis: Yes they can, I added a comment to say as much

brainiak/utils/fmrisim.py
CameronTEllis: Fantastic point. I had been testing this in cases where the event duration was greater than the TR (and so this isn't a problem). But as you point out when the events are shorter than a TR then you get this exact problem. I have shifted the downsampling step back to after convolution. I also changed the default temporal_resolution: it was 1000 so that it was measuring milliseconds, but you are right >100Hz is excessive

brainiak/utils/fmrisim.py
CameronTEllis: Rewrote this to clarify. The point is that the drift gets more basis functions (more kinks) when the run is longer

CameronTEllis: Thank you so much for the review again, these were fantastic comments and very insightful

brainiak/utils/fmrisim.py
```diff
@@ -1430,14 +1457,46 @@ def _generate_noise_temporal_drift(trs,
...
+
+            # What steps do you want to take for this basis function
+            timepoints_basis = (timepoints/basis_counter) + (np.random.rand()
+                                                             * np.pi * 2)
+
+            # Store the drift from this basis func
+            noise_drift[:, basis_counter - 1] = np.cos(timepoints_basis)
+
+        # Average the drift
+        noise_drift = np.mean(noise_drift, 1)
```
lcnature: Suggestion for future improvement: It is possible that the power of low frequency drift is higher than higher frequency drift. There may be some literature quantifying this property of fMRI data. It may be possible to weight fluctuations at different frequencies according to such pattern if it is reported in the literature, instead of putting equal weight. I am not expert on this. But I remember learning that fMRI power spectrum looks like 1/f.

tests/utils/test_fmrisim.py
```diff
@@ -97,6 +99,9 @@ def test_generate_stimfunction():
     signal_function = sim.convolve_hrf(stimfunction=stimfunction,
                                        tr_duration=tr_duration,
                                        )
+
+    max_response = np.where(signal_function != 0)[0].max()
+    assert 25 < max_response <= 30, "HRF is incorrect length"
```
lcnature: `is` -> `has`?

brainiak/utils/fmrisim.py
```diff
@@ -817,30 +841,27 @@ def convolve_hrf(stimfunction,
...
-        signal_function_temp = signal_function_temp[0::decimate_interval]
-
-        # Cut off the HRF
-        last_timepoint = stimfunction_temp.shape[0] / tr_duration
-        last_timepoint /= temporal_resolution
-        signal_function_temp = signal_function_temp[0:int(last_timepoint)]
+        # Down sample the stim function so that it only has one element per
+        # TR. This accelerates the convolution greatly
+        signal_vox = np.zeros((duration,))
+        for sample in list(range(duration)):
```
lcnature: I think the suggestion in this post might do the same as the for loop but may be more efficient.
https://stackoverflow.com/questions/15956309/averaging-over-every-n-elements-of-a-numpy-array
 Just be careful to make sure that the `reshape` works, i.e., the length of `signal_vox` is dividable by `stride`

mihaic:
The recommendation to use `nbstripout` is not automatically enforced. While we accepted non-stripped notebooks, we very much prefer not to pollute Git with non-code data, such as images. The long-term solution to using notebooks is something like `sphinx-gallery` (see PR #270).

tests/utils/test_fmrisim.py
CameronTEllis: Thanks!

brainiak/utils/fmrisim.py
CameronTEllis: Good suggestion, I will keep it in mind

brainiak/utils/fmrisim.py
```diff
@@ -473,7 +473,9 @@ def generate_stimfunction(onsets,
             # Check if the onset is more precise than the temporal resolution
             upsampled_onset = float(onset) * temporal_resolution
 
-            if upsampled_onset - np.round(upsampled_onset) != 0:
+            # Because of float precision, there can be issues. E.g.
+            #  float('1.001') * 1000 = 1000.99
+            if np.allclose(upsampled_onset, np.round(upsampled_onset)):
```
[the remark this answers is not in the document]
CameronTEllis: Good suggestion, I have made it a warning now so the user can decide. Thanks for all the help!

lcnature:
@CameronTEllis @mihaic I think the code looks pretty good to me. I checked the error message. It is from my test code of the `gen_design` function in 'utils`. I think it arises because in the past, the down-sampling after convolution in fmrisim was a direct down-sampling, while it is an averaging in a time window in the new code. I included that original test code based on the old fmrisim implementation. I was not to sure exactly which approach match the MRI physics better so I asked mrihelp yesterday. Mark told me that the acquisition time of a slice (ignoring multi-band) is the TR divided by the number of slices. So for any single slice, it seems that the old approach of direct down-sampling makes sense. But for slice-timing corrected image, interpolation is involved to infer the BOLD signal at one particular time for all slices. To me, this would be a bit close to the average signal within that TR, but not exactly. So I am still not sure whether direct down-sampling or averaging within a TR is better for your code -- both make some senses. Maybe it is up to you to choose one of them. If you stick with averaging, then I think deleting Lines 124-128 (https://github.com/brainiak/brainiak/blob/158e7b66e10b16710babadcb72186b9c542b1a63/tests/utils/test_utils.py#L124)  in `test/utils/test_utils.py` would resolve the error message. If you opt for direct down-sampling as you did before, then I think the error message will go away automatically.

mihaic:
@CameronTEllis, when implementing @lcnature's suggestion, please document your choice in a docstring.

CameronTEllis:
@lcnature,

This is an interesting nuance. As I see it, the contrast is between saying the signal within a TR is either represented by the activity at single time point or it reflects the average of the activity within a TR. Practically, these are small differences because the HRF is slow relative to the length of a TR. However, as you point out there are substantive theoretical differences. 

In an EPI we are indeed measuring voxel intensity in a slice by slice manner and this acquisition time is very brief. So in some sense the signal that a voxel represents should come from a single timepoint of the HRF. I believe the first approach is correct, but inaccurate if applied by downsampling as suggested.

If we downsampled by taking every nth timepoint, we would (re-)introduce a substantial bias (remember this is what we previously had as the default). For instance, if you have a signal function (estimate of the HRF convolved with the event sequence) that is 100,000 elements, and you want to down sample to 100 TRs, each of 1s duration, then you might take the 0th element, 100th element, 200th ... etc to represent your expected activity for each TR. This would effectively mean you are treating this and every voxel like it as something like the first acquired slice. You could pick a different nth element (the 50th or the 99th) but either way you are making an assumption that all signal is being acquired at the same time. This is also a problem because it can effectively shift the timing of events: sampling the 0th point is a TR different that sampling the 99th time point, so your peak response would come later in the 0th version than the 99th version.

As mentioned, we are in fact sampling data at multiple times throughout a TR. This means that taking the average of the signal function over that time window is approximately the same as if the voxels containing signal come from a distribution of voxels uniformly distributed across the slices. However, this is also often wrong, especially when we are doing whole brain acquisitions, since there is often non brain in the first and last slices. If we are simulating interleaved or multiband slice acquisition sequences this would be less of a problem but still present.

The most optimal way of designing the simulator would be to sample at a different time within a TR depending on the slice of the voxel being modelled. In other words, if the voxel is in the slice acquired first then the signal function should be sampled early in the TR, if the slice is last it should be sampled late in the TR. I could add this option to the code (provide a slice time of the voxel for each signal that is being simulated). This would be good additional functionality that would provide a lot of flexibility; although as mentioned before it would make little practical difference compared to averaging (I would default to take the middle time point).

If you agree with this logic I will go ahead and make this change.

Thanks,
Cameron

lcnature:
Hi @CameronTEllis 
Yes, I agree that for most of the usage, the two would not make much difference. The only case that they may differ a lot is when people use super long TR in some auditory task in order to make the auditory stimuli and the scanner noise far apart in time.

And yes, if the simulator capture the acquisition time difference between slices, it would be much closer to the reality. But of course this can be left for future extension if you want to do it.

But I think the worry about choosing to sample at 0th or 99th interval is less of concern. This is because if the simulator assume the same acquisition time for all slices, the user should make sure the simulated data are used to represent slice-timing corrected data. And in practice when doing slice-timing correction, you always specify which slice all other slices are aligned to. As long as the same timing choice is made for both the simulator and the real data, choosing the 0th or 99th slice are both correct.
The down-sampling here is different from the issue with down-sampling in the previous PR. Because last time the down-sampling was done to the event-timing series, in which sparse binary events may be entirely missed by the down-sampling. But here we are down-sampling at the simulated fMRI signal, which is already a smooth signal. (and that is why down-sampling and averaging are probably very close for most of the usage cases)

Maybe another (minor) argument for using direct down-sampling here is that respiratory or heart beat noise are sampled at discrete time instead of averaged within a TR in your code, if I understand correctly. Maybe it makes slightly more sense to use the same approach for both signal and noise.

Other people behind mrihelp just stopped by my office and some suggestion they gave include checking what other packages like FSL choose to do... They are happy to answer any questions you have, btw.

CameronTEllis:
@lcnature,

Thank you for the follow up. I think your suggestion to simulate the data 'as if' it is slice time corrected is a good one. FSL does slice time correction by aligning to the middle time point in a TR so if I downsample at the middle time point then this will resolve this. This is a much simpler solution than providing a slice time option.

You are right about the difference in downsampling the event time course rather than the convolved response, my mistake. 

Fortunately this will be a very minor change since the mean and middle time point are equal when the trend within a TR is linear (which is not always the case but is often close).

I will implement this now and re-commit the code. I will make sure to note this in the docstring

Thanks,
Cameron

tests/utils/test_utils.py
```diff
@@ -123,7 +123,7 @@ def test_gen_design():
         'gen_design does not treat missing values correctly')
     design5 = gen_design(stimtime_files=[files['FSL2']],
                          scan_duration=[48, 20], TR=1)
-    assert np.all(np.isclose(design4, design5[::2])), (
+    assert (design4 - design5[::2]).mean() < 0.1, (
```
lcnature: Is this still necessary given you have used direct down-sampling?
If it is still necessary. Maybe it should be the absolute value of difference that is smaller than some threshold, and 0.1 on the average difference may be too loose.

CameronTEllis: Up to you to decide whether it is still necessary.

I tested this value and it seemed in an appropriate range for the average difference. Do you have an alternative suggestion?

lcnature: Ah I see, I thought they will become exactly the same so `np.all(np.isclose...)` can actually work. But maybe I am wrong. 
Oh I think I know why, you are using the time of the middle slice. So then design4 and design5 are guaranteed to be different (they are off by 0.5 s). Maybe we should just remove this test between design4 and design5, given your current implementation.

CameronTEllis: Yes that is exactly right, the middle slice makes it different than if it was the first or last slice. I actually like this test though since they should be similar, just not the exact same. I vote to keep it in so that we know future changes to these functions don't have a big effect

lcnature: OK! Let's keep it. Does `assert (np.abs(design4 - design5[::2])).mean() < 0.1` work? I just think the mean should be applied to some metric of distance, rather than just the subtraction. If not, I think decreasing TR would make it work.

lcnature:
@mihaic The PR looks fine for me now! Thanks for your effort @CameronTEllis !

mihaic:
@lcnature, you have the right to squash and merge the commits to master yourself. First you have to update your review to formally approve the PR.

Issue #326, opened 2018-02-01
Unreliable coverage testing


mihaic:
There seems to be some variability in the coverage results reported. See discussion in PR #324.

A potential solution is changing the coverage testing to disallow lowering the coverage, not only to compare against a specified threshold.

mihaic:
The increase in coverage from PR #327 should get us over the threshold again, but it does not help with reliability.

Also, see this code comment in Coverage itself which indicates expected randomness:
https://bitbucket.org/ned/coveragepy/src/7b080c304b9219cdc33dd0b303db14d411c419fe/tests/test_concurrency.py?at=default&fileviewer=file-view-default#test_concurrency.py-362

PR #335, opened 2018-02-14
dev: Add coverage variation to deal with flakiness

mihaic:
One known source of flakiness is the interaction between Python
multiprocessing and Coverage.py:
https://bitbucket.org/ned/coveragepy/src/7b080c304b9219cdc33dd0b303db14d411c419fe/tests/test_concurrency.py?at=default&fileviewer=file-view-default#test_concurrency.py-362

See also issue #326.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,1 +1,6 @@
+coverage:
+    status:
+        project:
+            default:
+                threshold: 1
 comment: false


gdoubleyew:
Mihai, I'm not familiar with codecov, but looking at the docs it says "threshold: allowed to drop X% and still result in a "success"".
So by adding threshold 1, are you allowing 1% test failures to still be a success?
You mention occasional multiprocess test flakiness do to some randomness, do you use the @flaky decorator?

mihaic:
Thanks for having a look at this PR, @gdoubleyew.

Codecov is not looking at directly at test results. In fact, it only runs if the tests were successful. So the threshold only allows the test coverage to drop, it does not allow tests to fail.

The flakiness does not cause test failures in BrainIAK, as far as I can tell. Therefore, we do not need the `@flaky` decorator to pass tests. Instead, the flakiness causes the coverage to vary. Using `@flaky` with `min_passes > 1` will decrease the chance of a drop in coverage. However, even with a high `min_passes`, there is still a chance that coverage appears to drop, so we would still see spurious failures. Therefore, I think it is best to introduce the threshold.

PR #337, opened 2018-02-16
Add explanation for the assumption of what counts as signal in BRSA, correct default optimizer of BRSA

lcnature:
add notes in the documentation for the different assumption in BRSA from traditional RSA on what counts as signal. Also corrected the mismatch between documentation and actual implementation of the default optimizer. L-BFGS-B appears to require less memory and fits faster in each iteration than BFGS. But it is possible that more iterations are needed because Hessian is approximated, so the default max number of iteration is increased[](url)


mihaic:
@lcnature, we are preparing a new release to incorporate PR #339. Will you address the comments (they are really simple) today or tomorrow to get this PR in?

lcnature:
@mihaic Yes! I will resubmit today :)

PR #338, opened 2018-02-18
add latest htfa paper

xiazhu:


mihaic:
@xiazhu, thanks for keeping the docs up to date.

Note that your master branch will now be conflicting with the BrainIAK master branch because we use squash commits. That is why we recommend creating branches for PRs as opposed to working on the master branch.

PR #341, opened 2018-03-02
initial commit of matrix normal base API , regression, and MN-RSA

mshvartsman:
Massive squashed commit of matrix-normal functionality. Includes the base API, matrix-normal RSA, matrix-normal regression, and an example. Have a separate commit ready to go for matrix-normal SRM but this is already huge so let's do this first. 

Paging @narayanan2004 for double checking I didn't mess up with what I picked out of the private repo, @lcnature for reviewing RSA.


mihaic:
@narayanan2004, @lcnature, are you in touch with @mshvartsman about this PR?

lcnature:
@mihaic Sorry, not yet. I will have a look at it after Friday. @mshvartsman I just had a quick glance a the console output of the automatic testing system, it seems that there are quite a few formatting complaint in brainiak/utils/utils.py and a few in your brainiak/matnormal module. You can probably see them if you just run the run-checks.sh

mshvartsman:
Github rendering of math in notebooks seems pretty bad unfortunately -- it looks fine locally.

mshvartsman:
Sorry for so many vacuous commits as I get this ship-shape: for some reason `run-checks.sh` gives different results on my machine vs on the CI box.

mshvartsman:
Current linux failures look to be due to tensorflow being incompatible with old libc (e.g. https://github.com/tensorflow/tensorflow/issues/177), I'm guessing because the test server runs Centos/Springdale 6. There are various hackarounds possible (easiest might be installing TF from conda) but I don't know how to make any of them work with the CI setup. Any advice?

mihaic:
@mshvartsman, you are right. TensorFlow does appear to require glibc >= 2.17 (https://github.com/tensorflow/tensorflow/issues/53), although this requirement is not explicitly documented, as far as I can tell.

We do want to be able to depend on TensorFlow, so I am looking into a workaround for our Jenkins server.

mihaic:
I added PR #357 as a workaround for the TensorFlow issue.

mshvartsman:
@lcnature As mentioned on slack, the garbled math rendering is an issue with github's jupyter notebook renderer -- it looks fine locally. I don't know of a workaround -- @mihaic do you?

mihaic:
After PR #357 is merged, I will modify the Jenkins configuration to pass the flag. The Jenkins configuration is not part of this repo. Travis is not affected.

As for the GitHub Latex rendering, I am not aware of a workaround, but I don't think it matters. Users can see the rendering locally, as you mentioned; additionally, we should use our HTML docs for displaying examples (see issue #268).

mihaic:
@mshvartsman, please add the commit from PR #357 to your branch, so we can use the Jenkins workaround.

mihaic:
I didn't mean to be obscure. Merging master will add the commit.

mihaic:
Actually, I had modified the Jenkins config. Investigating...

mihaic:
There are now two build issues: the master build is failing (unrelated to this PR, maybe related to PR #357) and this PR build is still failing with glibc errors. Suggestions are welcome. I will investigate more later today or tomorrow.

mshvartsman:
Am happy to try to troubleshoot but I think I need credentials for Jenkins. Also: should we roll back #357 just in case so master build stops failing?

mihaic:
@mshvartsman, thanks for the offer to help. I suggest you make PRs in your own fork which should trigger Travis builds without any special authorization. See mine: https://github.com/mihaic/brainiak/pull/7

mihaic:
I think I found the culprit for the master breakage: Pip! Version 10, recently released, handles the installation of Pybind11 differently.

Updating Pip is the only difference between this passing commit:
https://travis-ci.org/mihaic/brainiak/builds/367886270

And this failing commit:
https://travis-ci.org/mihaic/brainiak/builds/367890418

I am looking into fixing this now. Then I will go back to adding support for TensorFlow on Jenkins.

mihaic:
New versions are fun! I'm experiencing yet another bug on my machine:
https://github.com/pypa/pip/issues/5224

mihaic:
I propose we do not use Pip 10 for the moment. See PR #358.

mihaic:
The master build issue is fixed so I am back to the TensorFlow issue. I am facing this Conda vs Pip issue:
https://github.com/conda/conda/issues/5969

mihaic:
I am struggling to make progress. CentOS 6, TensorFlow, Conda, Pip, and Environment Modules do not play well together. I will reach out to the server admins to see if they are willing to upgrade to CentOS 7. If not, we should consider abandoning Jenkins for the moment and relying only on Travis.

lcnature:
@mshvartsman sounds great!

mihaic:
The usage of TensorFlow (required by this PR) is blocked by PR #359.

brainiak/matnormal/utils.py
```diff
@@ -0,0 +1,73 @@
+import functools  # https://danijar.com/structuring-your-tensorflow-models/
```
mihaic: What is the role of this link? Did the idea come from there? Are you using code from there?

mihaic:
If you merge master, the Linux test on Jenkins should pass.

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,234 @@
...
+
+    # precompute sigma_row^{-1} * x
+    solve_row = row_cov.Sigma_inv_x(x)
+    logdet_row = row_cov.logdet
+
+    return _mnorm_logp_internal(colsize, rowsize, logdet_row,
+                                logdet_col, solve_row, solve_col)
+
+
+def matnorm_logp_marginal_row(x, row_cov, col_cov, marg, marg_cov):
```
lcnature: I don't know the math behind the following functions. But the structure of the codes seem fine.

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+           'CovDiagonalGammaPrior',
+           'CovUnconstrainedCholesky',
+           'CovUnconstrainedCholeskyWishartReg',
+           'CovUnconstrainedInvCholesky',
+           'CovKroneckerFactored']
+
+
+class CovBase:
+    """Base metaclass for noise covariances
+    """
```
mshvartsman: As per @lcnature, I should provide a little more info about how ABCs work and how devs should subclass from it, etc.

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+    def logdet(self):
+        """ log|Sigma|
+        """
+        pass
+
+    @abc.abstractmethod
+    def Sigma_inv_x(self, X):
+        """Given this Sigma and some X, compute :math:`Sigma^{-1} * x`
+        """
+        pass
```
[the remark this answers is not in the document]
mshvartsman: As per discussion, I will try to get rid of the scope stuff altogether in favor of `Reuse=True` or at worst, properties.

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+        return tf.exp(self.log_sigma)
+
+    def get_optimize_vars(self):
+        """ Returns a list of tf variables that need to get optimized to fit
+            this covariance
+        """
+        return [self.log_sigma]
+
+    @define_scope
+    def logdet(self):
```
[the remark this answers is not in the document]
mshvartsman: correct, either should call it log_sigmasq or should add a factor of 2 (and verify rest of math w.r.t. this)

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+       in terms of its precision cholesky
+    """
+
+    def __init__(self, size, invSigma=None):
+        if invSigma is None:
+            self.Linv_full = tf.Variable(tf.random_normal([size, size],
+                                         dtype=tf.float64), name="Linv_full")
+        else:
+            self.Linv_full = tf.Variable(np.linalg.cholesky(invSigma),
+                                         name="Linv_full")
```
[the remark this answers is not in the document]
mshvartsman: In general I think we need some basic standard on parameterization of things, so I may put that in docs.

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Use matrix determinant lemma for determinant:
+    .. math::
+    \log|(\Sigma + AQA')| = \log|Q^{-1} + A' \Sigma^{-1} A|
+    + \log|Q| + \log|\Sigma|
+    """
+
+    # we care about condition number of i_qf
+    if logging.getLogger().isEnabledFor(logging.DEBUG):
+        A = tf.Print(A, [_condition(Q.Sigma_inv + tf.matmul(A,
```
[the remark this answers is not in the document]
mshvartsman: alternative here is to eval the object and do the logging -- happy to do either?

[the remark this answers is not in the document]
mihaic: Right, we should not introduce a new output method, unless there is something wrong with our current logging.

brainiak/matnormal/covs.py
mihaic: Because we are using Python 3, you should subclass from `abc.ABC`.

Regarding providing more info, a note that the class has abstract methods with a link to the Python library docs (https://docs.python.org/3/library/abc.html#abc.abstractmethod) should be enough.

tests/matnormal/test_matnormal_logp_conditional.py
```diff
@@ -0,0 +1,89 @@
...
+# X is m x n, so A sould be m x p
+
+m = 5
+n = 4
+p = 3
+
+rtol = 1e-7
+
+
+def test_against_scipy_mvn_col_conditional():
```
lcnature: I am not sure if the naming of col_conditional of this function and the test for matnorm_logp_conditional_row is intentional.

tests/matnormal/test_matnormal_rsa.py
```diff
@@ -0,0 +1,75 @@
...
+
+    spacecov_model = CovIdentity(size=n_V)
+    timecov_model = CovDiagonal(size=n_T)
+
+    model_matnorm = MNRSA(time_cov=timecov_model,
+                          space_cov=spacecov_model)
+
+    model_matnorm.fit(tr['Y'], tr['X'])
+
+    RMSE = np.mean((model_matnorm.C_ - cov2corr(tr['U']))**2)**0.5
```
lcnature: It seems that this line should compare `model_matnorma.U_` to `tr['U']`

brainiak/matnormal/regression.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # @define_scope
+    def logp(self):
+        """ Log likelihood of model (internal)
+        """
+        y_hat = tf.matmul(self.X, self.beta)
+        resid = self.Y - y_hat
+        return matnorm_logp(resid, self.time_cov, self.space_cov)
+
+    def fit(self, X, y):
```
szorowi1: The optional arguments in the docstring (e.g. voxel_pos, times) are not inputs in the def statement. Is this by design? (Is CovGP not implemented?)

brainiak/matnormal/regression.py
```diff
@@ -0,0 +1,155 @@
...
+    ..math::
+    Y \\sim \\mathcal{MN}(X\beta, time_cov, space_cov)
+
+    Parameters
+    ----------
+    time_cov : subclass of CovBase
+        TR noise covariance class following CovBase interface.
+    space_cov : subclass of CovBase
+        Voxel noise covariance class following CovBase interface.
+    learnRate : real, default=0.01
```
szorowi1: Apologies if I'm missing something, but input argument learnRate doesn't seem to be an argument in the __init__ below. Do we need documentation for optimizer method and optCtrl?

mihaic:
@mshvartsman, I think @szorowi1 can provide more detailed feedback. I can imagine you are short on time these days. When do you anticipate you will be able to address feedback?

mshvartsman:
I think looking like mid-late July before I can seriously work on this: still working on basic life stuff like having my furniture delivered, and have travel coming up in a week or so. But I will commit to sitting down with @szorowi1 in person when he's here at the UW neuro summer school to work through this.

mihaic:
Excellent!

mihaic:
@mshvartsman, @szorowi1, how's the summer school going? Are you getting the chance to work on this PR?

mihaic:
Good to hear from you, @mshvartsman. We can keep this PR open now that we know what you're up to. :)

mshvartsman:
It's in process :). I've collected 1/2 of the necessary signatures, hopefully soon.

lcnature:
Thanks!

brainiak/matnormal/utils.py
mshvartsman: It's code from there (or rather, from the gist linked from there). Gist is https://gist.github.com/danijar/8663d3bbfd586bffecf6a0094cd116f2, original website is down but per Wayback machine https://web.archive.org/web/20181116012513/https://danijar.com/structuring-your-tensorflow-models/ it is licensed CC BY-SA 3.0 , citation:

```@misc{hafner2016scopedecorator,
  author = {Hafner, Danijar},
  title = {Structuring Your TensorFlow Models},
  year = {2016},
  howpublished = {Blog post},
  url = {https://danijar.com/structuring-your-tensorflow-models/}
}```

Is that usable if we cite appropriately in that file?

File: brainiak/matnormal/__init__.py [hunk 12]
```diff
@@ -0,0 +1,254 @@
...
+     \\Tr\\left[\\C\\inv(\\X-\\M)\\trp\\R\\inv(\\X-\\M)\\right]
+
+
+Here :math:`X` and :math:`M` are both :math:`m\\times n` matrices, :math:`\\R`
+is :math:`m\\times m` and :math:`\\C` is :math:`n\\times n`.
+
+The `brainiak.matnormal` package provides structure to infer models that
+can be stated in the matrix-normal notation that are useful for fMRI analysis.
+It provides a few interfaces. `MatnormModelBase` is intended to be subclasses
+from by matrix-variate models. It provides a wrapper for the tensorflow
```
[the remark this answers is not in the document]
mshvartsman: -> "is intended as a base class for matrix-variate models", will revise.

brainiak/matnormal/utils.py
```diff
@@ -0,0 +1,73 @@
...
+import tensorflow as tf
+from scipy.stats import norm
+from numpy.linalg import cholesky
+
+
+def rmn(rowcov, colcov):
+    # generate random draws from a zero-mean matrix-normal distribution
+    Z = norm.rvs(norm.rvs(size=(rowcov.shape[0], colcov.shape[0])))
```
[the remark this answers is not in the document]
mshvartsman: Nope, typo. Good catch.

brainiak/matnormal/utils.py
```diff
@@ -0,0 +1,73 @@
...
+    return(cholesky(rowcov).dot(Z).dot(cholesky(colcov)))
+
+
+def doublewrap(function):
+    """
+    A decorator decorator, allowing to use the decorator to be used without
```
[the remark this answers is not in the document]
mshvartsman: See https://web.archive.org/web/20181116012513/https://danijar.com/structuring-your-tensorflow-models/. This is basically some fancy python that allows the function to create the TF graph node when it is first called and then return a handle to the same node in future times rather than create a new node every time things like the logp node gets called. I think the right thing to do here is to pull the danijar code into a separate file and make it clear this is not written by me.

brainiak/matnormal/utils.py
```diff
@@ -0,0 +1,73 @@
...
+    parentheses if not arguments are provided. All arguments must be optional.
```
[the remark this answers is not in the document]
mshvartsman: Yes, but will leave as-is since this is not my code.

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,234 @@
+import tensorflow as tf
+from .utils import scaled_I
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def _condition(X):
```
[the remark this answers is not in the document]
mshvartsman: Computes condition number, used for debugging and logging. Will add comment.

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,234 @@
...
+    s = tf.svd(X, compute_uv=False)
+    return tf.reduce_max(s)/tf.reduce_min(s)
+
+
+def solve_det_marginal(x, sigma, A, Q):
+    """
+    Use matrix inversion lemma for the solve:
+    .. math::
+    (\Sigma + AQA')^{-1} X =\\
```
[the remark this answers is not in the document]
mshvartsman: Good catch.

brainiak/matnormal/matnormal_likelihoods.py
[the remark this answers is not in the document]
mshvartsman: When this was written tensorflow logging wasn't very good :). Will rewrite using tf logging ops and hopefully this will just work.

brainiak/utils/utils.py
```diff
@@ -812,3 +813,307 @@ def p_from_null(X, two_sided=False,
         p = 1 - max_null_ecdf(real_data)
 
     return p
+
+
+def tf_solve_lower_triangular_kron(L, y):
+    """ Tensor flow function to solve L x = y
```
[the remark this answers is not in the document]
mshvartsman: Good catch.

brainiak/utils/utils.py
```diff
@@ -812,3 +813,307 @@ def p_from_null(X, two_sided=False,
...
+
+    """
+    n = len(L)
+    if n == 1:
+        return tf.matrix_triangular_solve(L[0], y)
+    else:
+        x = y
+        na = L[0].get_shape().as_list()[0]
+        n_list = tf.stack([tf.to_double(tf.shape(mat)[0]) for mat in L])
+        n_prod = tf.to_int32(tf.reduce_prod(n_list))
```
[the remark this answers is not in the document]
mshvartsman: This is @narayanan2004's code. But I believe some of these variables need to be ints because we use them for indexing later.

mshvartsman:
All signed, working on it now. Will take a little while but please feel free to bug me if there's no forward progress

brainiak/matnormal/utils.py
mihaic: I'm afraid we cannot use the reuse code under CC BY-SA 3.0, because that would require we also use CC BY-SA 3.0 for BrainIAK (it's a copylef license). I specifically asked long time ago in the Gist comments for permission to reuse the code under the Apache License 2.0, but got no response:
https://gist.github.com/danijar/8663d3bbfd586bffecf6a0094cd116f2#gistcomment-1982431

Maybe you should ask again, maybe via email, before giving up on reusing the code?

lcnature:
@mshvartsman Just curious if there is any update on this PR:)

brainiak/matnormal/covs.py
mshvartsman: done.

brainiak/matnormal/covs.py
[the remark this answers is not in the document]
mshvartsman: define_scope and similar stuff is gone now

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+        """Given this Sigma and some X, compute :math:`Sigma^{-1} * x`
+        """
+        return tf.cholesky_solve(self.L, X)
+
+    @define_scope
+    def logdet(self):
+        return 2 * tf.reduce_sum(tf.log(tf.matrix_diag_part(self.L)))
+
+
+class CovIdentity(CovBase):
```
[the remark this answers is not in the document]
mshvartsman: done (though now it's _cov since it's mostly not meant for external access)

brainiak/matnormal/utils.py
mshvartsman: I think this is gone now, resolving.

brainiak/matnormal/__init__.py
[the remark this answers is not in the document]
mshvartsman: resolved in new commit.

brainiak/matnormal/utils.py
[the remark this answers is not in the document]
mshvartsman: this is completely gone now, resolving.

brainiak/matnormal/utils.py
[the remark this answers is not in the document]
mshvartsman: gone now, resolving.

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,234 @@
...
+    \Sigma^{-1} - \Sigma^{-1} A (Q^{-1} +
+    A' \Sigma^{-1} A)^{-1} A' \Sigma^{-1}
+
+    Use matrix determinant lemma for determinant:
+    .. math::
+    \log|(\Sigma + AQA')| = \log|Q^{-1} + A' \Sigma^{-1} A|
+    + \log|Q| + \log|\Sigma|
```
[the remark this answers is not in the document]
mshvartsman: From https://en.wikipedia.org/wiki/Matrix_determinant_lemma#Generalization, let me know if this makes sense.

brainiak/matnormal/utils.py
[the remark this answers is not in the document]
mshvartsman: fixed in new commit, resolving.

brainiak/matnormal/matnormal_likelihoods.py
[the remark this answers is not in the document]
mshvartsman: Added comment in upcoming commit (and also caution that this impl is only well-defined for symmetric positive definite matrices).

brainiak/utils/utils.py
```diff
@@ -812,3 +813,307 @@ def p_from_null(X, two_sided=False,
...
+    x_s1 = tf.matrix_triangular_solve(L_masked, y_masked,
+                                      lower=lower, adjoint=adjoint)
+    x = tf.scatter_nd(maskindex, x_s1, tf.to_int64(tf.shape(y)))
+    return x
+
+
+def tf_solve_lower_triangular_masked_kron(L, y, mask):
+    """ Tensor flow function to solve L x = y
+    where L = kron(L[0], L[1] .. L[n-1])
+    and L[i] are the lower triangular matrices
```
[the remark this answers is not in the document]
mshvartsman: fixed.

brainiak/utils/utils.py
```diff
@@ -812,3 +813,307 @@ def p_from_null(X, two_sided=False,
...
+def tf_solve_upper_triangular_masked_kron(L, y, mask):
+    """ Tensor flow function to solve L^T x = y
+    where L = kron(L[0], L[1] .. L[n-1])
+    and L[i] are the lower triangular matrices
+
+    Arguments
+    ---------
+    L : list of 2-D tensors
+        Each element of the list must be a tensorflow tensor and
+        must be a lower triangular matrix of dimension n_i x n_i
```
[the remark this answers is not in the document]
mshvartsman: default to lower=True in TF, https://www.tensorflow.org/api_docs/python/tf/linalg/triangular_solve.

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+
+class CovAR1(CovBase):
+    """AR1 covariance
+    """
+    def __init__(self, size, rho=None, sigma=None, scan_onsets=None):
+
+        # Similar to BRSA trick I think
+        if scan_onsets is None:
+            self.run_sizes = [size]
+            self.offdiag_template = tf.constant(scipy.linalg.toeplitz(np.r_[0,
```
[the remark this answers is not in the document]
mshvartsman: going to leave this as an open issue once we merge the initial thing

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+        sigma = tf.exp(self.log_sigma)
+        return (self._identity_mat - rho * self.offdiag_template + rho**2 *
+                self.diag_template) / tf.square(sigma)
+
+
+class CovIsotropic(CovBase):
+    """Scaled identity (isotropic) noise covariance.
+    """
+
+    def __init__(self, size, sigma=None):
```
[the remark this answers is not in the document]
mshvartsman: yep, added

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+        super(CovIsotropic, self).__init__(size)
+        if sigma is None:
+            self.log_sigma = tf.Variable(tf.random_normal([1],
+                                         dtype=tf.float64), name="sigma")
+        else:
+            self.log_sigma = tf.Variable(np.log(sigma), name="sigma")
+
+    @define_scope
+    def sigma(self):
```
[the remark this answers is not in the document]
mshvartsman: this is gone, resovling

brainiak/matnormal/covs.py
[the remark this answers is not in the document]
mshvartsman: switched to make this explicitly `log_sigmasq`

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+    @define_scope
+    def logdet(self):
+        """ log|Sigma|
+        """
+        return self.size * self.log_sigma
+
+    def Sigma_inv_x(self, X):
+        """Given this Sigma and some X, compute :math:`Sigma^{-1} * x`
+        """
+        return X / self.sigma
```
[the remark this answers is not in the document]
mshvartsman: switched to `self.sigmasq` for clarity

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+
+    @define_scope
+    def Sigma_inv(self):
+        """ Sigma^{-1}.
+        """
+        return tf.diag(tf.ones([self.size], dtype=tf.float64)) / self.sigma
```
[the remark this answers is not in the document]
mshvartsman: this is gone, resovling

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+        """ Returns a list of tf variables that need to get optimized to fit
+            this covariance
+        """
+        return [self.logprec]
+
+    @define_scope
+    def logdet(self):
+        """ log|Sigma|
+        """
+        return -tf.reduce_sum(self.logprec)
```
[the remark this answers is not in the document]
mshvartsman: if `sigma` -> `diag_var` then this is correctly precision and I think the rest makes sense.

brainiak/matnormal/covs.py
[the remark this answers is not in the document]
mshvartsman: actually wait, I use it in `solve_det_marginal` and `solve_det_conditional` which are used extensively. Going to keep this comment open to remind myself to refactor this to more sensible naming convention.

brainiak/matnormal/covs.py
```diff
@@ -0,0 +1,542 @@
...
+        """ Returns a list of tf variables that need to get optimized to fit
+             this covariance
+        """
+        return [self.L_full]
+
+    @define_scope
+    def logdet(self):
+        """ log|Sigma| using a cholesky solve
+        """
+        return 2 * tf.reduce_sum(tf.log(tf.matrix_diag_part(self.L)))
```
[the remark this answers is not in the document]
mshvartsman: yeah, nice.

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+
+    # we care about condition number of i_qf
```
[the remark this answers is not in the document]
mshvartsman: revised comments here

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,234 @@
...
+
+    return solve, logdet
+
+
+def solve_det_conditional(x, sigma, A, Q):
+    """
+    Use matrix inversion lemma for the solve:
+    .. math::
+    (\Sigma - AQ^{-1}A')^{-1} X =\\
+    \Sigma^{-1} + \Sigma^{-1} A (Q -
```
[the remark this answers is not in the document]
mshvartsman: yep, fixed

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,234 @@
...
+    prod_term = tf.cholesky_solve(i_qf_cholesky, Atrp_Sinv)
+
+    solve = tf.matmul(sigma.Sigma_inv_x(scaled_I(1.0, sigma.size) +
+                      tf.matmul(A, prod_term)), x)
+
+    return solve, logdet
+
+
+def _mnorm_logp_internal(colsize, rowsize, logdet_row, logdet_col,
+                         solve_row, solve_col):
```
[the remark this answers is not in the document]
mshvartsman: added.

brainiak/matnormal/matnormal_likelihoods.py
mshvartsman: added.

brainiak/matnormal/mnrsa.py
```diff
@@ -0,0 +1,172 @@
...
+import logging
+
+__all__ = ['MNRSA']
+
+
+class MNRSA(BaseEstimator):
+    """ Matrix normal version of RSA.
+
+    The goal of this analysis is to find the covariance of the mapping from
+    some design matrixX to the fMRI signal Y. It does so by marginalizing over
```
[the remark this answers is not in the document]
mshvartsman: fixed

brainiak/matnormal/mnrsa.py
```diff
@@ -0,0 +1,172 @@
...
+    the actual mapping (i.e. averaging over the uncertainty in it), which
+    happens to correct a bias imposed by structure in the design matrix on the
+    RSA estimate (see Cai et al., NIPS 2016).
+
+    This implementation makes different choices about two things relative to
+    `brainiak.reprsimil.BRSA`:
+
+    1. The noise covariance is assumed to be kronecker-separable. Informally,
+    this means that all voxels has the same temporal covariance, and all time
+    points have the same spatialcovariance. This is in contrast to BRSA, which
```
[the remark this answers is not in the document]
mshvartsman: fixed

brainiak/matnormal/mnrsa.py
```diff
@@ -0,0 +1,172 @@
...
+    allows different temporal covariance for each voxel. On the other hand,
+    computational efficiencies enabled by this choice allow MNRSA to
+    support a richer class of space and time covariances (anything in
+    `brainiak.matnormal.covs`).
+
+    2. MNRSA does not estimate the nuisance timecourse X_0. Instead,
+    we expect the temporal noise covariance to capture the same property
```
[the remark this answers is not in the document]
mshvartsman: yep, updated doc

brainiak/matnormal/mnrsa.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # create a tf session we reuse for this object
+        self.sess = tf.Session()
+
+    def fit(self, X, y, structured_RSA_cov=None):
+        """ Estimate dimension reduction and cognitive model parameters
+
+        Parameters
+        ----------
+        X: 2d array
```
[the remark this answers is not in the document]
mshvartsman: yep, good catch. fixed

brainiak/matnormal/regression.py
mshvartsman: yeah, outdated docs, will fix this.

brainiak/matnormal/regression.py
mshvartsman: yeah, outdated docstring (covgp was excluded from this PR when I split things up from the private repo)

tests/matnormal/test_matnormal_logp_conditional.py
mshvartsman: oh yeah the names are swapped relative to what we're actually testing. fixed.

tests/matnormal/test_matnormal_rsa.py
mshvartsman: I'm computing RMSE on the correlation (`C_` vs `cov2corr(tr['U'])`), which is what fMRI people tend to care about -- why compare to cov instead? I'm super open to it, just wondering re: reasoning.

mshvartsman:
@szorowi1 @lcnature addressed a bunch of your comments and cleaned things up, please feel free to resolve anything that is resolved to your satisfaction / unresolve anything that I resolved prematurely. Major things still outstanding: 
- Logging fixes using new TF logging functionality. 
- Verifying that with ripping out all the `define_scope` stuff I haven't introduced TF graph inflation anywhere. 
- Investigating what to do about materialized cov and prec. Basically I use them in some of the linalg tricks for marginal/conditional solves and logdets in addition to testing, but materializing those would mean you need to hold a FCMA-style matrix in memory if you wanted to run whole-brain MNRSA.

mshvartsman:
In addition to these changes, some updates:
- Logging update: TF 1.0 API logging won't help much here: without eager mode I can log the tensor itself, not the value (which is useless) unless I have a session I can use to eval. This is fine for the logging calls inside the estimator class (which I've modified) but not in helper classes (all the matrix-normal linear algebra tricks). I don't think it's worth passing the session around just for better logging (though I'm happy to do that if you think that's a better option). I propose that for the current PR I keep old-school `tf.Print` logging (still gated behind log levels as it is now), and if the tool is starting to get usage there should be an eventual migration to TF 2.0  eager mode which would obviate some of these issues. 
- Scope stuff: I think we're good here also, and longer term this also completely goes away with TF 2.0. 
- Materialized cov and prec: I think again this could be a new issue and not part of the current PR, we ran covs up to 10K+ voxels for the paper which should be a good start for now. 

Based on this, I'd like to ask for a re-review and see if we can get this finally merged :).

lcnature:
Thanks @mshvartsman ! I still owe the review to Hejia's SRM PR, so I may not be able to review this very soon. Can @szorowi1 take a look perhaps?

mshvartsman:
Hi folks -- anyone get a chance to review this? Happy to get together on VC with @lcnature or @szorowi1 or whoever else to get this finally closed out.

setup.py
```diff
@@ -137,6 +137,8 @@ def finalize_options(self):
         'pybind11>=1.7',
         'psutil',
         'nibabel',
+        'typing',
+        'tensorflow<=1.15',  # brainiak.matnormal not compatible with tf2.0
```
mihaic, line 141: @mshvartsman, is there any chance to update the PR to use `tf.compat.v1` APIs such that it can be used with TensorFlow 2 without refactoring?

mshvartsman: At least some functionality didn't make it to tf.compat.v1 (e.g. inverse wishart and gamma distrs we use for priors), but let me assess how much work it'd be to reimplement or remove, stay tuned!

mshvartsman: Ok, here are the things without compat APIs as far as I can tell: 
1. Inverse-Wishart and Gamma priors for regularized covariances. I think we could possibly just remove those priors (I believe we used neither of them in the paper and it's probably not worth reimplementing over just moving to TF probability when we can). 
2. The `ScipyOptimizerInterface` we use to our problems into Scipy's solvers (usually L-BFGS) while handling session bookkeeping. Can't just cut this out -- can switch to something like Adam which would likely perform worse, or manage our own bookkeeping, or something else? Would appreciate advice on how to proceed here.

mihaic: Thanks for the quick investigation, @mshvartsman. Based on your findings, I think we should make compatibility with TensorFlow 2 a goal for merging the PR. I will try to get reviewers looking at it ASAP.

mshvartsman: (1) above is done. (2) has different options. The basic issue is that with most TF optimizers you can just pass a list of tensors and a loss and it knows how to optimize that loss w.r.t. those params -- this is how all the regular DL optimizers work (Adam, RMSProp, etc) and how `ScipyOptimizerInterface` worked in TF 1. For `scipy.optimize.minimize` and `tfp.optimizer.lbfgs_minimize` you need a function that takes theta and returns val and grad (f(theta) and f'(theta)). The current matrix-normal API doesn't really expect that, so the parameters are part of the various objects' states (model, cov etc) and spread out in random parts of the graph (e.g. structured covariance params, weights, etc). I see some options: 

1. Switch to one of the DL optimizers like Adam. It's a relatively easy change, but the optimizer will be worse than L-BFGS for our setting (no batches, no stochastic gradients). 
2. Write a bunch of kludgy code to take theta, unpack it, and inject it into the graph via `feed_dict` (if that's still a thing in TF 2). I think it'll work, but I'll need to be careful to make sure this always works no matter how people set up their models (it'd basically replicate the code in the old module I was using, https://github.com/tensorflow/tensorflow/blob/r1.15/tensorflow/contrib/opt/python/training/external_optimizer.py). 
3. Completely rewrite the API to be fully functional, so parameters will stop being parts of model state and we can just unpack them and pass them in. 

I don't like any of these options: are there any others? I'm not really a software guy :) so maybe there's something better to do here. Regardless, I'd like to get feedback from whoever will be reviewing (@mihaic / @lcnature / @szorowi1) before I get working on this, so we are on the same page on how to proceed.

mihaic: Good to hear about (1)! As far as I understand, @szorowi1 will be reviewing shortly. Let's address (2) after we hear from him.

szorowi1: Hi all, happy to help with review. I started looking through the code, but it's been long enough at this point I've somewhat lost the thread. Are there specific sections of the code I should be reviewing? Are there sections that have already been reviewed and I should leave be? And/or am most helpful be helping to address (2) above? Thanks!

mihaic: Thanks for having another look at the PR, @szorowi1. We do need help with (2), so if you have experience with the options proposed by @mshvartsman, please weigh in.

As for the review in general, please give your opinion in as much detail as you can, based on the time you can commit.

mshvartsman: Note that this was reviewed by @lcnature already, and I believe I have addressed all his comments, so hopefully should be close with the exception of the remaining TF2.0 changes.

brainiak/matnormal/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+    \\mathbf{X}_{ij} \\sim \\mathcal{MN}(\\mathbf{A}_{ij},
+    \\Sigma_{\\mathbf{X}i},\\Sigma_{\\mathbf{X}j})\\\\
+    \\mathbf{Y}_{jk} \\sim \\mathcal{MN}(\\mathbf{B}_{jk},
+     \\Sigma_{\\mathbf{Y}j},\\Sigma_{\\mathbf{Y}k})\\\\
+    \\mathbf{Z}_{ik}\\mid\\mathbf{X}_{ij},\\mathbf{Y}_{jk} \\sim
+     \\mathcal{MN}(\\mathbf{X}_{ij}\\mathbf{Y}_{jk} + \\mathbf{C}_{ik},
+      \\Sigma_{\\mathbf{Z}_i}, \\Sigma_{\\mathbf{Z}_k})\\\\
+
+
+We vectorize, and covert to a form we recognize as
```
lcnature, line 88: covert -> convert

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+def matnorm_logp_marginal_row(x, row_cov, col_cov, marg, marg_cov):
+    """
+    Log likelihood for marginal centered matrix-variate normal density.
+
+    .. math::
+        X \\sim \\mathcal{MN}(0, Q, C)\\
+        Y \\mid \\X \\sim \\mathcal{MN}(AX, R, C),\\
+        Y \\sim \\mathcal{MN}(0, R + AQA, C)
```
lcnature, line 223: Is it AQA or AQA^T?

brainiak/matnormal/mnrsa.py
```diff
@@ -0,0 +1,172 @@
...
+    For users: in general, if you are worried about voxels each having
+    different temporal noise structure,you should use
+    `brainiak.reprsimil.BRSA`. If you are worried about between-voxel
+    correlations or temporal covaraince structures that BRSA does not
+    support, you should use MNRSA.
+
+    .. math::
+        Y \\sim \\mathcal{MN}(0, \\Sigma_t + XLL^{\\top}X^{\\top}+
+         X_0X_0^{\\top}, \\Sigma_s)
+        U = LL^{\\top}
```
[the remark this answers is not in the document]
lcnature: Can you double check on this @mshvartsman ?

brainiak/matnormal/__init__.py
```diff
@@ -58,8 +58,8 @@
 
 The `brainiak.matnormal` package provides structure to infer models that
 can be stated in the matrix-normal notation that are useful for fMRI analysis.
-It provides a few interfaces. `MatnormModelBase` is intended to be subclasses
-from by matrix-variate models. It provides a wrapper for the tensorflow
+It provides a few interfaces. `MatnormModelBase` is intended is intended as a
```
lcnature, line 61: I think "is intended" is duplicated"

brainiak/matnormal/covs.py
```diff
@@ -3,438 +3,475 @@
...
-    def get_optimize_vars(self):
-        """ Returns a list of tf variables that need to get optimized to fit
-            this covariance
+    @property
+    def _prec(self):
+        """Expose the precision explicitly (mostly for testing /
+        visualization, materializing large covariances may be intractable)
         """
-        return []
+        return self.solve(tf.eye(self.size, dtype=tf.float64))
```
lcnature, line 73: I wonder if someone writes an inherent class and forgets to redefine `_prec`, will it take the default of inverse of this identity matrix without any warning?

brainiak/matnormal/covs.py
```diff
@@ -3,438 +3,475 @@
...
+                scipy.sparse.block_diag(sub_offdiags).toarray()
+            )
+            subdiags = [np.diag(np.r_[0, np.ones(r - 2), 0])
                         for r in self.run_sizes]
-            self.diag_template = tf.constant(scipy.sparse.
-                                             block_diag(subdiags)
-                                             .toarray())
+            self.diag_template = tf.constant(
+                scipy.sparse.block_diag(subdiags).toarray()
+            )
```
lcnature, line 162: Side note: I don't know how efficiently tensorflow handles these block diagonal matrices. If it is not very efficient, I wonder if (potentially in a future update), you would consider just separately computing the log likelihood for each run. Because basically there would be no temporal correlation between runs, and I think the total log likelihood is just the sum of each.

brainiak/matnormal/covs.py
```diff
@@ -3,438 +3,475 @@
...
+    and new noise sigma.
+
+    Parameters
+    ----------
+    size: int
+        size of covariance matrix
+    rho: float or None
+        initial value of autoregressive parameter (if None, initialize
+        randomly)
+    sigma: float or None
```
lcnature, line 131: It may be worth documenting here that it is actually the log of sigma.

brainiak/matnormal/covs.py
```diff
@@ -3,438 +3,475 @@
...
+                scipy.sparse.block_diag(subdiags).toarray()
+            )
 
         self._identity_mat = tf.constant(np.eye(size))
 
         if sigma is None:
-            self.log_sigma = tf.Variable(tf.random_normal([1],
-                                         dtype=tf.float64), name="sigma")
+            self.log_sigma = tf.Variable(
+                tf.random.normal([1], dtype=tf.float64), name="sigma"
```
lcnature, line 168: I wonder if the tf variable name should be "log_sigma" as well

brainiak/matnormal/covs.py
```diff
@@ -3,438 +3,475 @@
...
         if rho is None:
-            self.rho_unc = tf.Variable(tf.random_normal([1], dtype=tf.float64),
-                                       name="rho")
+            self.rho_unc = tf.Variable(
+                tf.random.normal([1], dtype=tf.float64), name="rho"
+            )
         else:
-            self.rho_unc = tf.Variable(np.log(rho), name="rho")
+            self.rho_unc = tf.Variable(
+                2 * tf.sigmoid(self.rho_unc) - 1, name="rho")
```
lcnature, line 179: I think self.rho_unc is not defined before this line, so it should not be an argument, right?
I wonder if you were thinking of tf.Variable(scipy.special.logit(rho / 2 + 0.5), name='rho_unc')

brainiak/matnormal/covs.py
```diff
@@ -3,438 +3,475 @@
...
+        """
+        # first, unconstrain rho and sigma
+        rho = 2 * tf.sigmoid(self.rho_unc) - 1
+        sigma = tf.exp(self.log_sigma)
+        # now compute logdet
+        return tf.reduce_sum(
+            input_tensor=2 * tf.constant(self.run_sizes, dtype=tf.float64) *
+            tf.math.log(sigma)
+            - tf.math.log(1 - tf.square(rho))
+        )
```
lcnature, line 193: I think since you already have a self.log_sigma, you don't need to take exp and log again.

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -1,61 +1,97 @@
...
+                    "Printing diagnostics for solve_det_marginal")
+        A = tf.compat.v1.Print(A, [_condition(Q._prec + tf.matmul(A, sigma.solve(A),
+                                                        transpose_a=True))],
+                     "lemma_factor condition")
+        A = tf.compat.v1.Print(A, [_condition(Q._cov)], "Q condition")
+        A = tf.compat.v1.Print(A, [_condition(sigma._cov)], "sigma condition")
+        A = tf.compat.v1.Print(A, [tf.reduce_max(input_tensor=A), tf.reduce_min(input_tensor=A)], "A minmax")
+
+    # cholesky of (Qinv + A' Sigma^{-1} A), which looks sort of like
+    # a schur complement by isn't, so we call it the "lemma factor"
```
lcnature, line 67: by -> but

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -64,147 +100,214 @@ def solve_det_conditional(x, sigma, A, Q):
...
-    Log likelihood for centered matrix-variate normal density. Assumes that
-    row_cov, col_cov, and marg_cov follow the API defined in CovBase.
+    Log likelihood for centered marginal matrix-variate normal density.
 
-    When you marginalize in mnorm, you end up with a covariance S + APA',
-    where P is the covariance of A in the relevant dimension.
+    .. math::
+        X \\sim \\mathcal{MN}(0, R, Q)\\
+        Y \\mid \\X \\sim \\mathcal{MN}(XA, R, C),\\
+        Y \\sim \\mathcal{MN}(0, R, C + AQA)
```
lcnature, line 262: A'QA?

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -64,147 +100,214 @@ def solve_det_conditional(x, sigma, A, Q):
...
 
-    return _mnorm_logp_internal(colsize, rowsize, logdet_row,
-                                logdet_col, solve_row, solve_col)
+    return _mnorm_logp_internal(
+        colsize, rowsize, logdet_row, logdet_col, solve_row, solve_col
+    )
 
 
 def matnorm_logp_conditional_row(x, row_cov, col_cov, cond, cond_cov):
     """
```
lcnature, line 297: I wonder if you would consider adding docstring for this function

brainiak/utils/kronecker_solvers.py
```diff
@@ -0,0 +1,314 @@
...
+    if n == 1:
+        return tf.linalg.triangular_solve(L[0], y)
+    else:
+        x = y
+        na = L[0].get_shape().as_list()[0]
+        n_list = tf.stack([tf.cast(tf.shape(input=mat)[0], dtype=tf.float64) for mat in L])
+        n_prod = tf.cast(tf.reduce_prod(input_tensor=n_list), dtype=tf.int32)
+        nb = tf.cast(n_prod/na, dtype=tf.int32)
+        col = tf.shape(input=x)[1]
+
```
lcnature, line 39: Not entirely understanding the logic below. But hopefully it is sound

brainiak/utils/kronecker_solvers.py
```diff
@@ -0,0 +1,314 @@
...
+                  [(na-i-1)*nb, col])
+            x = tf.concat(axis=0, values=[xt, xinb, xina])
+
+        return x
+
+
+def tf_solve_upper_triangular_kron(L, y):
+    """ Tensorflow function to solve L^T x = y
+    where L = kron(L[0], L[1] .. L[n-1])
+    and L[i] are the lower triangular matrices
```
lcnature, line 58: Since the name indicates upper triangular matrix, should L be lower or upper?

brainiak/utils/kronecker_solvers.py
```diff
@@ -0,0 +1,314 @@
...
+                  * tf.reshape(
+                      tf.tile(tf.reshape(t, [-1, 1]), [i, 1]), [i*nb, col]))
+            x = tf.concat(axis=0, values=[xt, xinb, xina])
+
+        return x
+
+
+def tf_kron_mult(L, x):
+    """ Tensorflow multiply with kronecker product matrix
+    Returs kron(L[0], L[1] ...) * x
```
lcnature, line 104: Returns

brainiak/utils/kronecker_solvers.py
```diff
@@ -0,0 +1,314 @@
...
+           tf.transpose(a=tf.reshape(tf.transpose(a=x), [-1, nb])))
+        y = tf.zeros_like(x)
+        for i in range(na):
+            ya, yb, yc = tf.split(y, [i*nb, nb, (na-i-1)*nb], 0)
+            yb = tf.reshape(tf.matmul(tf.reshape(xt, [nb*col, na]),
+                                      tf.transpose(a=tf.slice(L[0],
+                                                            [i, 0],
+                                                            [1, na]))),
+                            [nb, col])
+            y = tf.concat(axis=0, values=[ya, yb, yc])
```
lcnature, line 140: If I understand correctly, this function differs from the previous two functions in that when you calculate yb, it does not depend on ya and yc. So I wonder whether this repeated splitting and concatenation will unnecessarily claim at least twice as much as memory. I think tf.concat will add new node to the graph. If it does not just create a view of the concatenated sub-matrices, I worry that it will actually claim a memory of the size of y for every iteration of i. Do you know if this is the case? Perhaps an alternative is just to create an empty list before the loop. In every iteration you just create yb and append to the list. At the end you only do one concatenation outside the loop.

setup.py
mshvartsman: I think I have the TF 2.0 workaround for the above that does not change any of the APIs and works with scipy.minimize: https://github.com/mshvartsman/brainiak/tree/mnorm-eager. My plan is to resolve @lcnature's comments in this PR first, then get the TF 2.0 fixes in. I don't know if there's a way to stack PRs in github: should I just merge that branch into this one, or do something else?

mihaic: Yes, I would prefer not to have a TF<2 dependency even temporarily, so please add those changes to this PR.

brainiak/matnormal/covs.py
mshvartsman: Good point. Shall I write up an issue for this after merging? Prefer not to incorporate it into this PR, lots of things here already.

brainiak/utils/kronecker_solvers.py
mshvartsman: If I understand this correctly (@narayanan2004's code) this is correct: an upper triangular solve but the input shows up as lower triangular (meaning it solves `L^T x = y` for input `L` and `y` as in the docstring).

brainiak/utils/kronecker_solvers.py
mshvartsman: I could be missing something (this is also Narayanan's code), but in my understanding the `tf.split` call in every iteration is different (since it is a function of `i` the loop index). So on every iteration we make a different split, operate on the middle, and recombine. On every iteration of the loop, `ya` and `yc` should be different, so doing the split and concat outside will do the wrong thing. Here's a slightly less terse (matlab) implementation of what should be the same algo https://github.com/pillowlab/fastASD/blob/master/tools_kron/kronmult.m.

mshvartsman:
Ok, @lcnature's comments all fixed / incorporated except for the discussion above. So @mihaic if you're happy with everything except the TF2 support in https://github.com/mshvartsman/brainiak/tree/mnorm-eager, I can merge that into the PR and that should be the last piece.

setup.py
mshvartsman: Deleted previous comment, TF2.0 fix is still broken (I thought I had a clever solution but instead discovered how loose some of my tests are).

brainiak/matnormal/covs.py
mshvartsman: `self.solve` does `Sigma^-1 I` for the current Cov, and always has to be implemented (it's an abstract method). So basically, this is as-intended: you should rarely need to redefine `_cov` or `_prec`, since they're inefficient and mostly used for debugging, and they will always call the solve for you.

brainiak/matnormal/covs.py
mshvartsman: Yep, good catch.

brainiak/matnormal/covs.py
mshvartsman: (main current exception is in `CovAR1` where we redefine it, but that is because that's what you need for an efficient solve)

brainiak/matnormal/matnormal_likelihoods.py
mshvartsman: added

mshvartsman:
Just pushed the TF2.0 fixes, everything now works in eager mode and the example notebook even demonstrates how to set it up. I also removed the calls to `tf.compat.v1.print` since we can do normal logging now. None of the math / modeling changed and the APIs are basically the same, so the main thing to look at would be the new way that the `fit` methods look internally, plus the new utils that pack/unpack things for Scipy.

mihaic:
Excellent news, @mshvartsman! Thank you for the effort. @lcnature, what do you think?

mshvartsman:
@mihaic do you think you could take a look at the tooling changes? It seems like @lcnature has been mostly happy with the math/modeling side but it'd be nice to have an additional critical software-focused eye on `make_val_and_grad`, and the parameter packing/unpacking functions in `matnormal/utils.py` before we unleash them upon unsuspecting neuroscientists :). Also okay if you want to wait until I finish up the final docstrings and make Travis happy.

mshvartsman:
I'm stuck on Travis: I need TF 2.3 or greater for `tensorflow_probability`. If I don't put in any version requirement (previous commit), it defaults to 1.14 and fails when I try to import `tensorflow_probability`. If I try to force it to 2.3 (which is what `tensorflow_probability` says it needs), it fails because no installation candidate. I remember we had an issue like this before when the CI box ran an old Centos, is this a similar problem? `run-checks.sh` and all matnormal tests pass on my local machine, and I think all docstrings are cleaned up as well.

mihaic:
I get the same error on my Ubuntu machine, `Could not find a version that satisfies the requirement tensorflow>=2.3`. Looking into it.

mihaic:
Upgrading Pip solves the issue. Now to figure out how to use Pip>=10 to build BrainIAK...

mshvartsman:
BrainIAK builds fine for me with pip 20 and `--no-use-pep517`. That said, I run into another issue: SSSRM uses pymanopt, which isn't updated for TF2. If TF isn't installed, pymanopt just ignores it with a `try/except ImportError`. But with TF2 installed, it fails because `tf.Session` no longer exists. Per https://github.com/pymanopt/pymanopt/issues/86 pymanopt isn't fixing this any time soon. Is there a workaround?

mshvartsman:
Ok, I made a (really hacky) workaround where after we import pymanopt but before we instantiate our first `Problem` we tell pymanopt to pretend TF isn't installed.

mihaic:
I think the Pymanopt fix is fine. Based on my own experience with convincing Pymanopt to adopt changes and make subsequent releases, we should not wait for them.

Now after PR #479, Travis should find TensorFlow 2.3.

setup.py
```diff
@@ -138,6 +138,9 @@ def finalize_options(self):
         'pybind11>=1.7',
         'psutil',
         'nibabel',
+        'typing',
```
mihaic, line 141: Are we using features from `typing` that are not in Python 3.6? If not, the version in the standard library will work.

brainiak/utils/utils.py
```diff
@@ -18,6 +18,7 @@
 import psutil
 from .fmrisim import generate_stimfunction, _double_gamma_hrf, convolve_hrf
 from scipy.fftpack import fft, ifft
+
```
mihaic, line 21: Please remove the whitespace changes in this file.

setup.py
mshvartsman: I haven't put any type annotations in so I don't know how this made it in there (and actually, I think it may have been breaking some of the travis builds?). Fixed.

brainiak/matnormal/matnormal_likelihoods.py
```diff
@@ -0,0 +1,429 @@
...
+    logging.log(
+        logging.DEBUG,
+        f"lemma_factor condition={lemma_cond}",
+    )
+    logging.log(logging.DEBUG, f"Q condition={_condition(Q._cov)}")
+    logging.log(logging.DEBUG, f"sigma condition={_condition(sigma._cov)}")
+    logging.log(
+        logging.DEBUG,
+        f"sigma max={tf.reduce_max(input_tensor=A)}," +
+        "sigma min={tf.reduce_min(input_tensor=A)}",
```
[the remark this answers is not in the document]
mshvartsman: Good catch, thank you. Fixed.

brainiak/utils/utils.py
mshvartsman: Oops, another good catch -- thanks. Fixed.

brainiak/matnormal/covs.py
```diff
@@ -343,49 +357,43 @@ def __init__(self, size=None, Sigma=None):
...
+        Cholesky factor of this covariance
+        """
+        return unflatten_cholesky_unique(self.L_flat)
 
-        # We save a log here by using the diag of L_full
-        return 2 * tf.reduce_sum(input_tensor=(tf.linalg.diag_part(self.L_full)))
+    @property
+    def logdet(self):
+        return 2 * tf.reduce_sum(input_tensor=tf.math.log(
+                                 tf.linalg.diag_part(self.L)))
```
lcnature, line 385: One question: is there any possibility that the diagonal values are negative? The covariance can still be positive definite if L has negative diagonal value. How about you take the square of the diagonal part before log and remove the multiplication by 2 outside?

brainiak/matnormal/covs.py
```diff
@@ -427,43 +435,57 @@ class CovUnconstrainedInvCholesky(CovBase):
...
+        """
+        Inverse of Cholesky factor of this covariance
+        """
+        return unflatten_cholesky_unique(self.Linv_flat)
 
     @property
     def logdet(self):
-        return -2 * tf.reduce_sum(input_tensor=tf.math.log(tf.linalg.diag_part(self.Linv)))
+        return -2 * tf.reduce_sum(
+            input_tensor=tf.math.log(tf.linalg.diag_part(self.Linv))
```
lcnature, line 477: Same concern about negative value in Linv.

brainiak/matnormal/covs.py
```diff
@@ -511,66 +533,79 @@ def __init__(self, sizes, Sigmas=None, mask=None):
...
+                 for mat in self.L]
+            )
             n_prod = tf.reduce_prod(input_tensor=n_list)
             logdet = tf.stack(
-                [tf.reduce_sum(input_tensor=tf.math.log(tf.linalg.tensor_diag_part(mat)))
-                 for mat in self.L]
+                [
+                    tf.reduce_sum(
+                        input_tensor=tf.math.log(
+                            tf.linalg.tensor_diag_part(mat))
```
lcnature, line 579: same concern, if the one above is valid.

brainiak/matnormal/covs.py
```diff
@@ -511,66 +533,79 @@ def __init__(self, sizes, Sigmas=None, mask=None):
...
             mask_reshaped = tf.reshape(self.mask, n_list)
             logdet = 0.0
             for i in range(self.nfactors):
                 indices = list(range(self.nfactors))
                 indices.remove(i)
-                logdet += tf.math.log(tf.linalg.tensor_diag_part(self.L[i])) * tf.cast(
-                    tf.reduce_sum(input_tensor=mask_reshaped, axis=indices), dtype=tf.float64
-                )
+                logdet += (tf.math.log(tf.linalg.tensor_diag_part(self.L[i])) *
+                           tf.cast(
```
lcnature, line 594: same here

brainiak/matnormal/covs.py
mshvartsman: `self.L` is a property that is computed by running `unflatten_cholesky_unique(self.L_flat)` (line 376), defined on line 74 of `utils.py`, as such:
```
    L = tfp.math.fill_triangular(L_flat)
    # exp diag for unique parameterization
    L = tf.linalg.set_diag(L, tf.exp(tf.linalg.diag_part(L)))
```
So basically, I exp the diagonal to make it positive, which makes the parameterization unique and valid (I believe). 

I'm open to another operation that makes it positive (like squaring) if you think that's a better idea, but I prefer not to remove the factor of 2 outside even is squaring the diag would be mathematically valid -- I think operating in log space will generally be more numerically stable, and having it in the `logdet` method makes it easy to understand what's going on here (since it looks like a textbook formula).

brainiak/matnormal/covs.py
mshvartsman: Same response -- also defined via unique cholesky parameterization.

brainiak/matnormal/covs.py
mshvartsman: Also defined via unique cholesky parameterization, should be valid.

brainiak/matnormal/covs.py
mshvartsman: Also defined via unique cholesky parameterization, should be valid.

mshvartsman:
Thank you for your comments @lcnature, I have verified that in all places where we do logdet for cholesky the cholesky is defined in a unique way.

brainiak/matnormal/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    return tf.linalg.tensor_diag(tf.ones([size], dtype=tf.float64) * x)
+
+
+def flatten_cholesky_unique(L):
+    """
+    Flattens nonzero-elements Cholesky (triangular) factor
+    into a vector, and logs diagonal to make parameterization
+    unique. Inverse of unflatten_cholesky_unique.
+    """
+    L[np.diag_indices_from(L)] = np.log(np.diag(L))
```
lcnature, line 69: This change is in-place. I wonder if there is a risk of unintentionally change variable (L) outside the function here? Is it better to make a copy of L before changing the diagonal?

lcnature:
@mshvartsman @mihaic Other than the last comment above, I think the PR is good to go!

brainiak/matnormal/covs.py
lcnature: Agree. Sorry I forget you had this way to ensure unique parametrization. So this is fine.

brainiak/matnormal/utils.py
mshvartsman: I ended up replacing the numpy calls with TF calls for consistency. The new call (https://www.tensorflow.org/api_docs/python/tf/linalg/set_diag) returns a tensor the same as the input tensor but with a changed diagonal, i.e. it is a copy.

brainiak/utils/utils.py
```diff
@@ -964,7 +964,6 @@ def array_correlation(x, y, axis=0):
     r : float or 1D ndarray
         Pearson correlation values for input variables
     """
-
```
mihaic, line 967: Please remove this one whitespace difference as well.

mshvartsman: I think that's done.

mshvartsman:
The conda build fix doesn't look too onerous, we might as well do it. I think I got all the spots where pymanopt is special-cased and did the same for TFP? The one place I'm not sure about is `jenkins-server-customization`: does that still need to be there, or does that go away with the hackaround I've got in SSSRM? 

Latest TF in conda is 2.2 though and TFP requires 2.3 so we can either do the pypi workaround or just wait until TF2.3 makes it into the conda repo, up to you.

lcnature:
@mihaic @mshvartsman Looks good to me too! I am happy to approve! Thank you both for contributing and managing!

lcnature (approved):
Just hit the approve button here

mihaic:
@lcnature, thanks a lot for helping with this monumental PR!

@mshvartsman, since you are up fixing the Conda build, could you please try TensorFlow from PyPI as well? Please ignore `jenkins-server-customization`; we decided to stop testing on Jenkins recently.

mshvartsman:
Tried it, not sure what happened with the build there. Looks like TFP successfully installed without complaining about missing TF2.3 but then it couldn't find all the TF libs and froze. Any ideas?

.conda/build.sh
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
 
-# Install pymanopt and tensorflow_probability via pip because there isn't a conda package
+# Install pymanopt and tensorflow_probability via pip because there isn't a conda 
+# package, and install tensorflow 2.3 from pip because tensorflow_probability
+# requires it and latest in conda is 2.2. 
+PIP_NO_INDEX=False $PYTHON -m pip install tensorflow>=2.3
```
mihaic, line 6: You need to quote or escape the `>`, otherwise Bash interprets it as a redirect, e.g., `tensorflow\>=2.3`. We cannot see in the Travis logs what happened to the installation because of this redirect.

mshvartsman: Ah, thanks. I blame my fancy programming font ligature.

mshvartsman: I don't think that was it though? Still fails.

mihaic: Thanks for trying. Not sure what is happening. I'll experiment with building the Conda package manually to understand. If I don't figure it out in the next few days, I plan to merge the PR nevertheless and solve the build issue later.

mihaic (approved):

.conda/build.sh
mshvartsman: Sounds good! I'll get rolling on the next piece (MN-SRM) and PR that when it's ready.

mihaic:
The good news is that there is a TensorFlow 2.3 package in Conda (tensorflow-base). The bad news is that it is not compiled for macOS, only Linux.

mshvartsman:
Does that change our plans for merging?

mihaic:
My comment was a sort of explanation for the failure of commit 5e472a1. Still planning to merge for the upcoming release and still looking for the best way to build the Conda package.

mihaic:
@mshvartsman, I have a new proposal in light of the difficulty in packaging TensorFlow 2.3 for macOS with Conda. What if I make a new `brainiak-matnormal` package? It would be a pure Python package, so we will be able to publish a binary wheel for it on PyPI. That would be easily installable using either Pip or Conda, with no prerequirement from the operating system.

I also thought about using Setuptools `extras_require` to make `matnormal` an optional part of the BrainIAK package, but that does not work in Conda as far as I can tell.

mihaic:
@mshvartsman, another solution I see is to use [`extras_require`](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies) and do nothing in Conda. Instead, we would instruct Conda users to install the TensorFlow 2.3 separately. The instructions would be:

No TensorFlow, `matnormal` cannot be used:
> pip install brainiak
or
> conda install brainiak

TensorFlow installed, `matnormal` can be used:
> pip install brainiak[matnormal]
or
> conda install brainiak && pip install -r requirements-matnormal.txt

What do you think?

mihaic:
@mshvartsman, to make some progress, I will implement the `extras_require` proposal. We can change it later.

mihaic:
@mshvartsman, thank you very much for your patience! I hope the optional dependencies approach will work well.

Thank you for your help, @narayanan2004, @lcnature, and @szorowi1.

PR #343, opened 2018-03-15
add p_from_null_mem_saving

takua624:
move some calculations to ISFC (or ISC) function to save memory

### DIFF
--- a/brainiak/utils/utils.py
+++ b/brainiak/utils/utils.py
@@ -795,3 +795,55 @@ def p_from_null(X, two_sided=False):
         p = 1 - max_null_ecdf(X[..., 0])
 
     return p
+
+def p_from_null_mem_saving(X, two_sided=False, max_null=0, min_null=0): # add 2 parameters
+    """Compute p value of true result from null distribution
+    Given an array containing both a real result and a set of null results,
+    computes the fraction of null results larger than the real result (or,
+    if two_sided=True, the fraction of null results more extreme than the real
+    result in either the positive or negative direction).
+    Note that all real results are compared to a pooled null distribution,
+    which is the max/min over all null results, providing multiple
+    comparisons correction.
+    Parameters
+    ----------
+    X : ndarray with arbitrary number of dimensions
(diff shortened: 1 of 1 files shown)


buildbot-princeton:

buildbot-princeton:

mihaic:
Jenkins, OK to test.

brainiak/utils/utils.py
```diff
@@ -795,3 +795,55 @@ def p_from_null(X, two_sided=False):
...
+    # Compute maximum/minimum in each null dataset
+    # max_null = np.max(X[..., 1:], axis=leading_dims) # deleted
+    # min_null = np.min(X[..., 1:], axis=leading_dims) # deleted
+
+    # Compute where the true values fall on the null distribution
+    max_null_ecdf = ecdf(max_null)
+    if two_sided:
+        min_null_ecdf = ecdf(min_null)
+        # p = 2 * np.minimum(1 - max_null_ecdf(X[..., 0]), min_null_ecdf(X[..., 0])) # deleted
+		p = 2 * np.minimum(1 - max_null_ecdf(X), min_null_ecdf(X)) # added
```
mihaic: You are changing the dimensions of the result by omitting the slicing. You should only apply `ecdf` to the real results.

brainiak/utils/utils.py
```diff
@@ -795,3 +795,55 @@ def p_from_null(X, two_sided=False):
         p = 1 - max_null_ecdf(X[..., 0])
 
     return p
+
+def p_from_null_mem_saving(X, two_sided=False, max_null=0, min_null=0): # add 2 parameters
```
mihaic: There is no need for such comments. We have Git to show us changes.

brainiak/utils/utils.py
takua624: In the original code, max_null_ecdf take X[...,0] as input, which is a 3-dimensional array (voxel*voxel*subject), where X itself is 4-dimensional.
Since in the memory-saving version, X will be 3-dimensional, shouldn't the dimension of p the same as the original version?

PR #344, opened 2018-03-15
modify ISFC

takua624:
1. modify p_from_null to take a bool parameter "memory_saving"
2. modify ISFC so that the voxel-by-voxel-by-subject array generated during each iteration of num_perm is not stored anymore. ISFC should always use the memory_saving mode of p_from_null


buildbot-princeton:

mihaic:
Jenkins, OK to test.

mihaic:
@takua624, if this supersedes your previous PRs, please close them.

takua624:
OK, thanks for the notice, I closed them.

mihaic:
Thanks.

You should consider running `pr-check` yourself before pushing to GitHub to speed up the testing. Even better, look at our contribution guide and run the checkers are tests individually:
http://brainiak.org/docs/contributing.html#how-to-contribute

takua624:
I don't know if it's the problem with the setting, some the Git commands listed in the how-to-contribute page resulted in "fatal" error message, and I can't figure out why.
So I'm using the web-browser based GitHub to make pull requests. Sorry for the inconvenience.

brainiak/isfc.py
```diff
@@ -84,7 +84,6 @@ def isc(D, collapse_subj=True, return_p=False, num_perm=1000,
         n_perm = num_perm
     else:
         n_perm = 0
-
     ISC = np.zeros((n_vox, n_subj, n_perm + 1))
```
manojneuro: This still has the n_perm parameter in the array. Is this okay?

takua624: Yes, it is. I want to make sure the modification on ISFC works before also changing ISC. But, yeah, I can also change ISC now

brainiak/utils/utils.py
```diff
@@ -773,25 +774,58 @@ def p_from_null(X, two_sided=False):
         above the null) or two-sided (testing for both significantly positive
```
cbaldassano: You should update this docstring to indicate that X can either have this form, or contain only the real result if max/min arguments are used

brainiak/utils/utils.py
```diff
@@ -773,25 +774,58 @@ def p_from_null(X, two_sided=False):
...
+        else:
+            p = 1 - max_null_ecdf(X[..., 0])
+
     else:
-        p = 1 - max_null_ecdf(X[..., 0])
+        # maximum & minimum in each null dataset should be provided as input
+        max_null = max_null_input
+        min_null = min_null_input
+
+        max_null_ecdf = ecdf(max_null)
```
cbaldassano: This should be refactored to avoid repeating these near-identical lines. I would just compute max/min if necessary (and then set X = X[...,0]), then compute the p value with a single shared block of code

brainiak/utils/utils.py
```diff
@@ -773,25 +774,58 @@ def p_from_null(X, two_sided=False):
...
+            p = 2 * np.minimum(1 - max_null_ecdf(X),
+                               min_null_ecdf(X))
+            p = np.minimum(p, 1)
+        else:
+            p = 1 - max_null_ecdf(X)
 
     return p
+
+
+def gen_null_array(n_subj, n_perm, collapse_subj, float_type):
```
cbaldassano: This is pretty specific to the eventseg code, so I would move this to eventseg.py. It also needs a (brief) docstring, and maybe a better name to clarify that the arrays are empty rather than null samples (e.g. initialize_null_arrays)

brainiak/isfc.py
```diff
@@ -80,37 +86,50 @@ def isc(D, collapse_subj=True, return_p=False, num_perm=1000,
...
-    if return_p:
-        n_perm = num_perm
-    else:
-        n_perm = 0
-
-    ISC = np.zeros((n_vox, n_subj, n_perm + 1))
+    n_perm = num_perm*int(return_p)
+    max_null, min_null = gen_null_array(n_subj, n_perm,
+                                        collapse_subj, float_type)
+    ISC = np.zeros((n_vox, n_subj), dtype=float_type)
```
cbaldassano: Both here and in ISFC below, it is not obvious to me why you need separate code for the real and permuted computations - in the old version of the code, I just had a single loop that computed the real value on the first iteration, and then all the permutations.

brainiak/isfc.py
```diff
@@ -80,37 +86,50 @@ def isc(D, collapse_subj=True, return_p=False, num_perm=1000,
...
+                                              subj[v, :])[0]
+    if collapse_subj:
+        ISC = np.mean(ISC, axis=1)
 
-    for p in range(n_perm + 1):
+    for p in range(n_perm):
+        # Randomize phases of D to create next null dataset
+        D = phase_randomize(D, random_state)
         # Loop across choice of leave-one-out subject
+        tmp_ISC = np.zeros((n_vox, n_subj), dtype=float_type)
```
cbaldassano: In ISFC below, you compute this average iteratively for each subject (so the memory usage doesn't scale with the number of subjects) - you could do the same here, though it doesn't matter much since ISC is much less memory-intensive.

brainiak/isfc.py
```diff
@@ -160,30 +185,48 @@ def isfc(D, collapse_subj=True, return_p=False,
...
+        tmp_ISFC = (tmp_ISFC+tmp_ISFC.T)/2
+        ISFC[:, :, loo_subj] = tmp_ISFC
+    if collapse_subj:
+        ISFC = np.mean(ISFC, axis=2)
+
+    for p in range(n_perm):
+        # Randomize phases of D to create next null dataset
+        D = phase_randomize(D, random_state)
         # Loop across choice of leave-one-out subject
+        ISFC_mean = np.zeros((n_vox, n_vox), dtype=float_type)
```
cbaldassano: I would rename this variable to make it clear that this is a null value.

brainiak/utils/utils.py
takua624: Thanks, I added some descriptions to the docsting

brainiak/utils/utils.py
takua624: Thanks. I added docstrings for this function.
But I can't find "eventseg.py", should I add it to /eventseg/event.py ?
But event.py contains a class, wouldn't it be strange to add this function to the class?

brainiak/utils/utils.py
takua624: Thanks, I merged the near-identical lines

brainiak/isfc.py
takua624: Thanks, I renamed it to ISFC_null

brainiak/utils/utils.py
cbaldassano: My apologizes, that was a typo on my part (I was commenting on my own separate pull request at the same time) - I meant to say that this is specific to the *IS(F)C* code, so it should be moved to isfc.py. The utils file should be for functions that are expected to be useful in multiple modules, but I don't think this one is.

tests/utils/test_utils.py
```diff
@@ -227,8 +227,17 @@ def test_p_from_null():
...
     p_1side = p_from_null(X, two_sided=False)
     assert np.isclose(p_1side, [0.25, 1]).all(), "One-sided p value incorrect"
 
     p_2side = p_from_null(X, two_sided=True)
     assert np.isclose(p_2side, [0.5, 0]).all(), "Two-sided p value incorrect"
+
+    p_2side_m = p_from_null(Y, two_sided=True,
+                            max_null_input=Y_max,
+                            min_null_input=Y_min)
+    assert np.isclose(p_2side, p_2side_m).all(), "new p_null differs!"
```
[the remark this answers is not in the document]
takua624: Thanks, modified accordingly

brainiak/utils/utils.py
takua624: I see. But I removed this function, so there's no problem now.

brainiak/isfc.py
```diff
@@ -68,6 +68,12 @@ def isc(D, collapse_subj=True, return_p=False, num_perm=1000,
         A random number generator instance to define the state of the
         random permutations generator.
 
+    float_type : either float16, float32, or float64,
+        depending on the required precision,
+        and available memory in the system.
+        cast all the arrays generated during the execution to
+        specified float type, in order to save memor
```
mihaic: Typo: "memor". Also, please write regular sentences in the documentation, i.e., starting with a capital, full stop only at the end.

brainiak/isfc.py
```diff
@@ -80,37 +86,50 @@ def isc(D, collapse_subj=True, return_p=False, num_perm=1000,
...
-        ISC = np.mean(ISC, axis=1)
+                tmp_ISC[v, loo_subj] = stats.pearsonr(group[v, :],
+                                                      subj[v, :])[0]
+        if not collapse_subj:
+            max_null[p] = np.max(tmp_ISC)
+            min_null[p] = np.min(tmp_ISC)
+        if collapse_subj:
+            tmp_ISC = np.mean(tmp_ISC, axis=1)
+            max_null[p] = np.max(tmp_ISC)
+            min_null[p] = np.min(tmp_ISC)
```
mihaic: We try not to repeat code. How about:
```
if collapse_subj:
    tmp_ISC = np.mean(tmp_ISC, axis=1)
max_null[p] = np.max(tmp_ISC)
min_null[p] = np.min(tmp_ISC)
```

brainiak/isfc.py
```diff
@@ -145,6 +164,12 @@ def isfc(D, collapse_subj=True, return_p=False,
         A random number generator instance to define the state of the
         random permutations generator.
 
+    float_type : either float16, float32, or float64
+        depending on the required precision
+        and available memory in the system.
+        cast all the arrays generated during the execution to
+        specified float type, in order to save memory
```
mihaic: Same comment about writing regular sentences.

brainiak/utils/utils.py
```diff
@@ -767,31 +768,45 @@ def p_from_null(X, two_sided=False):
...
+        If max_null_input and min_null_input are provided,
+        X should contain only the real result
 
     two_sided : bool, default:False
         Whether the p value should be one-sided (testing only for being
         above the null) or two-sided (testing for both significantly positive
         and significantly negative values)
 
+    max_null_input
+    min_null_input
```
mihaic: Please follow the parameter doctring format: specify the type on the same line as the parameter name.

To avoid repeating the description, you could say: "See `max_null_input`."

To reference the ISFC module, you should write "see `brainiak.isfc`" (with backticks). This will automatically create a link in the HTML docs, because we use ":py:obj:" as our default Sphinx role:
http://www.sphinx-doc.org/en/stable/domains.html#role-py:obj

Finally, remember sentence capitalization.

brainiak/isfc.py
takua624: Thanks! Didn't notice that when I realized the max_null should always be a 1-D array. Fixed.

brainiak/utils/utils.py
takua624: Thank you, fixed accordingly.

brainiak/isfc.py
takua624: Thank you, fixed accordingly.

brainiak/isfc.py
takua624: Thank you, fixed accordingly.

mihaic:
@takua624, we only allow merges for PRs that are up-to-date with BrainIAK master to avoid errors. Therefore, you must merge the latest changes from BrainIAK master into your master branch. You can do this using the "Update branch" button at the end of the PR page:
https://github.com/brainiak/brainiak/pull/344

Note that after we merge the PR, your master branch will now be conflicting with the BrainIAK master branch, because we use squash commits to keep a clean history. That is why we recommend creating branches for PRs as opposed to working on the master branch.

mihaic:
Thanks again for you contribution, @takua624!

If you have ideas about making contributing easier, please open issues or reach out on Gitter or the email list.

PR #353, opened 2018-03-26
Eventseg updates

cbaldassano:
Making transition matrix more flexible, and improving utility functions and tests

### DIFF
--- a/brainiak/eventseg/event.py
+++ b/brainiak/eventseg/event.py
@@ -18,12 +18,12 @@ segment the timeseries into events with stable activity patterns. After
 learning the signature activity pattern of each event, the model can then be
 applied to other datasets to identify a corresponding sequence of events.
 
-Full details are available in the bioRxiv preprint:
+Full details are available in:
 Christopher Baldassano, Janice Chen, Asieh Zadbood,
 Jonathan W Pillow, Uri Hasson, Kenneth A Norman
 Discovering event structure in continuous narrative perception and memory
 Neuron, Volume 95, Issue 3, 709 - 721.e5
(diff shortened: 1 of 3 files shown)


brainiak/eventseg/event.py
```diff
@@ -60,14 +60,12 @@ class EventSegment(BaseEstimator):
...
+        in the order they appear in the array
+
     Attributes
     ----------
-    p_start, p_end: length n_events+1 ndarray
-        initial and final prior distributions over events
-
-    P: n_events+1 by n_events+1 ndarray
-        HMM transition matrix
-
```
mihaic: I am not sure if someone was really using these attributes, but please add a "removal" news fragment announcing their removal.

brainiak/eventseg/event.py
```diff
@@ -60,14 +60,12 @@ class EventSegment(BaseEstimator):
     n_iter: int : default 500
         Maximum number of steps to run during fitting
 
+    event_chains: ndarray with length = n_events
+        Array with unique value for each separate chain of events, each linked
+        in the order they appear in the array
```
mihaic: As far as I can tell, this is the first time the term "chain" is used. I think it would help to provide a definition.

brainiak/eventseg/event.py
cbaldassano: I actually decided to revert this change and keep these attributes publicly exposed, especially since the P matrix is now more complicated and so it could help users to debug.

brainiak/eventseg/event.py
cbaldassano: Good idea, I added more info to the top of the file

PR #357, opened 2018-04-09
Add TensorFlow workaround for our Jenkins server

mihaic:
We want to be able to use TensorFlow as a dependency. TensorFlow
currently requires glibc >= 2.17 (not formally documented, see
https://github.com/tensorflow/tensorflow/issues/53#issuecomment-156575907).
We are currently running Jenkins on CentOS 6, which has glibc < 2.17.
Installing TensorFlow with Conda also installs the required glibc.

### DIFF
--- a/pr-check.sh
+++ b/pr-check.sh
@@ -113,6 +113,23 @@ $activate_venv $venv || {
 python3 -m pip install $ignore_installed -U -e . || \
     exit_with_error_and_venv "Failed to install BrainIAK."
 
+# Allow installation on OSes we do not formally support by installing packages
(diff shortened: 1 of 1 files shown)


mihaic:
This is how I tested on the Jenkins server:
```bash
conda create -n tmp python=3
source activate tmp
python3 -m pip install tensorflow
python3 -c "import tensorflow"  # fail
conda install --yes tensorflow
python -c "import tensorflow"  # success
```

mihaic:
Thanks for reviewing, @mshvartsman. Merged.

PR #364, opened 2018-04-26
Created reconstruction method

dhuberdeau:
Added a reconstruct folder, including Inverted Encoding Models (1.Brouwer, G. J. & Heeger, D. J. Decoding and Reconstructing Color from Responses in Human Visual Cortex. J. Neurosci. 29, 13992–14003 (2009).).

TODO: Complete test functions and example function for the inverted encoding model method.


buildbot-princeton:

buildbot-princeton:

mihaic:
Jenkins, add to whitelist.

mihaic:
@dhuberdeau, are you making progress? Please reach out if you need help.

dhuberdeau:
@mihaic, thanks for reaching out. Yes, getting to these changes this week. I'm sure I'll have some questions, so I'll be in touch.

dhuberdeau:
This is a pull request for the Inverted Encoding Model (IEM) implementation. It creates a new folder, reconstruct, to which additional methods of reconstruction can be added.

This particular IEM implementation is for 1-D data only. Future development should add multi-dimensional reconstruction capabilities.

I'm not sure who would be appropriate as a reviewer. Perhaps Qi?

-David

mihaic:
@dhuberdeau, do you mean @qihongl?

dhuberdeau:
Hi @mihaic: Yes. I heard through the grapevine that @qihongl might be a good reviewer for this. I haven't asked him yet, though. Do you have any suggestions?

mihaic:
Let's hear from him first. @qihongl?

qihongl:
Hi, Folks! Sorry for the delayed response! I will start working on this by the end of next week!

dhuberdeau:
Thanks Qihong! I'll be available to make changes as you see fit. Don't hold back..

qihongl:
@dhuberdeau @mihaic 
I haven't look at the detail, but just a quick question: should we follow the standard code comment style? 
e.g., SRM: https://github.com/brainiak/brainiak/blob/master/brainiak/funcalign/srm.py

mihaic:
@qihongl, indeed. We actually formalized the coding standards in the docs:
http://brainiak.org/docs/contributing.html#standards

qihongl:
@mihaic Yep. Thanks for the reference!

mihaic:
@qihongl, any progress with the review?

@dhuberdeau, the "WIP" part of the PR title is not necessary anymore, right?

qihongl:
@mihaic Oops, I haven't started. I was waiting for the documentation and then I forgot. I will start working on it today or tomorrow! 

@dhuberdeau Have you started documenting the code? Any documentation would be really helpful.

dhuberdeau:
Hi @mihaic and @qihongl,  I'll add in proper documentation and do another commit. I'll let you know when that is done and ready for review. -David

qihongl:
@dhuberdeau awesome! Thanks!

CameronTEllis:
@mihaic Not sure about the Ruby error on Travis. Sounds like it is failing on BRSA?

dhuberdeau:
Thanks @qihongl, not sure why the build is getting stuck, but it seems like it's a time out issue, and I don't think that has to do with the code.

mihaic:
I don't think the Travis "Ruby" job failure is related to this PR either (that is actually the Conda package job). Please ignore it for know. The other Travis and Jenkins jobs are sufficient.

qihongl:
Got it! Thanks! @CameronTEllis @dhuberdeau @mihaic

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,290 @@
...
+
+    W_: sklearn.linear_model model containing weight matrix that
+        relates estimated channel responses to response amplitude
+        data
+    """
+    def __init__(self, n_channels=5, range_start=0, range_stop=180):
+
+        self.n_channels = n_channels  # default = 5
+        self.range_start = range_start  # in degrees, 0 - 360, def=0
+        self.range_stop = range_stop  # in degrees, 0 - 360, def=180
```
qihongl: I guess it would be nice to also do input validation for `range_start` and `range_stop`?

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,290 @@
...
+            raise ValueError("Insufficient channels.")
+        # Check that there is enough data.. should be more
+        # samples than voxels (i.e. X should be tall)
+        shape_data = np.shape(X)
+        shape_labels = np.shape(y)
+        if len(shape_data) < 2:
+            raise ValueError("Not enough data")
+        else:
+            if np.size(X, 0) <= np.size(X, 1):
+                raise ValueError("Data Matrix ill-conditioned")
```
qihongl: I guess if the error message says something more direct, like "the number of observations should be larger than the number of voxels," it might be easier for the user to understand what's going on?

dhuberdeau: Yes, good suggestion. I'll change that...

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,290 @@
...
+        for parameter, value in parameters.items():
+            setattr(self, parameter, value)
+        return self
+
+    def _define_channels(self):
+        """Define basis functions (aka channels).
+
+        Parameters
+        ----------
+            self.
```
qihongl: Do we need to list `self` in the list of input params?

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,290 @@
...
+"""
+
+# Authors: David Huberdeau (Yale University) &
+# Peter Kok (Yale University), 2018
+
+import logging
+import numpy as np
+from sklearn import linear_model
+from sklearn.base import BaseEstimator
+import math
```
qihongl: Since the only usage of `math` is `math.pi`, I would recommend using `np.pi`, so that you don't need to import `math` at all. The following stackoverflow QA says that they are the same: https://stackoverflow.com/questions/12645547/should-i-use-scipy-pi-numpy-pi-or-math-pi

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,290 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Inverted Encoding Model (IEM)
+
+    Method to decode and reconstruct features from data.
+
+    The implementation is roughly based on the following publications:
```
qihongl: I think it would be useful to briefly comment on what's the difference between this implementation and what's proposed in the paper.

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,290 @@
...
+        iem: self.
+        """
+        # Check that there are channels specified
+        if self.n_channels < 2:
+            raise ValueError("Insufficient channels.")
+        # Check that there is enough data.. should be more
+        # samples than voxels (i.e. X should be tall)
+        shape_data = np.shape(X)
+        shape_labels = np.shape(y)
+        if len(shape_data) < 2:
```
qihongl: Should `len(shape_data) == 2`?

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,290 @@
...
+        else:
+            if np.size(X, 0) <= np.size(X, 1):
+                raise ValueError("Data Matrix ill-conditioned")
+            if shape_data[0] != shape_labels[0]:
+                raise ValueError(
+                    "Mismatched data samples and label samples")
+
+        self.C_, self.C_D_ = self._define_channels()
+        n_train = len(y)
+        F = np.empty((n_train, self.n_channels))
```
qihongl: I think this section can be benefited from additional comments. I guess this is trying to find the best channel for each training example to construct a design matrix?

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,290 @@
...
+            pred_dir: numpy array of estimated feature values.
+        """
+        # Check that there is enough data.. should be more
+        # samples than voxels (i.e. X should be tall)
+        shape_data = np.shape(X)
+        if len(shape_data) < 2:
+            raise ValueError("Not enough data")
+        else:
+            if np.size(X, 0) <= np.size(X, 1):
+                raise ValueError("Data Matrix ill-conditioned")
```
qihongl: similar thoughts -> line 127, 131

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,290 @@
...
+            channels[i, :] = np.cos(np.linspace(0, math.pi, channel_density)
+                                    - shifts[i]) ** channel_exp
+        # Check that channels provide sufficient coverage
+        ch_sum_range = np.max(np.sum(channels, 0)) - min(np.sum(channels, 0))
+        if ch_sum_range > np.deg2rad(self.range_stop - self.range_start)*0.1:
+            # if range of channel sum > 10% channel domain size
+            raise ValueError("Insufficient channel coverage.")
+        return channels, channel_domain
+
+    def _predict_channel_responses(self, X):
```
qihongl: Is this function necessary? It seems that the only role of `_predict_directions()` is to call this function, so I was wondering if it is more efficient to combine `_predict_channel_responses()` with `_predict_directions()`.

brainiak/reconstruct/iem.py
mihaic: We should not include `self`. See the following guideline referenced in our contributing docs:
"Do not include the `self` parameter in the ``Parameters`` section."

brainiak/reconstruct/iem.py
dhuberdeau: Yes. I changed this, but also changed the comments to indicate that the values are more flexible. It now checks just that range_start is less than range_stop.

brainiak/reconstruct/iem.py
dhuberdeau: Good calls. I made the message more verbose, and also added a check that the matrix is not ill conditioned. Good suggestion with bumpy.linalg.cond.

brainiak/reconstruct/iem.py
dhuberdeau: got rid of any of those.

brainiak/reconstruct/iem.py
dhuberdeau: Excellent suggestion. Changed that.

brainiak/reconstruct/iem.py
dhuberdeau: Yes, I added a description.

brainiak/reconstruct/iem.py
dhuberdeau: Good idea. I changed that.

brainiak/reconstruct/iem.py
dhuberdeau: I added extra comments.

brainiak/reconstruct/iem.py
dhuberdeau: Good idea. I changed this.

brainiak/reconstruct/iem.py
dhuberdeau: This is a good suggestion, however, I wanted to keep the possibility, however remote, that someone might want to see what the channel responses are independently of just the prediction based on them. I noticed one of my lab mates doing so (although I think it was just a sanity check). If it's alright, I left them separate.

dhuberdeau:
Hi @qihongl, Thanks for the great reviews! They will really improve the code. I changed all but one of your suggestions (see above), and just committed the new files. Hopefully everything runs smoothly (it did on my machine, as far as I could tell). 

Keep those reviews coming - if you see anything else, let me know and I'll see what I can do. 

Thanks,
David

brainiak/reconstruct/iem.py
qihongl: I see! Thanks! That makes sense!

qihongl:
@dhuberdeau Great! Thank you very much! 

I just went through `iem_example_illconditioned_X.ipynb` and `iem_example_synthetic_data.ipynb`, they look great! I just some have minor suggestions. For `iem_example_synthetic_data.ipynb`, I think it would be helpful to add axis labels, titles, etc to the plots. I would also move some longer comments from the code blocks to markdown blocks. 

I have a question about `iem_example_synthetic_data.ipynb`. I tried some other choices of `n` and `dim`. And I noticed that when `n` is larger (or smaller), say `n=900`, the method didn't work well. Do you know why is this the case? I thought larger `n` should make decoding easier. 

And what is `iem_example_openfmri.ipynb`? Are you still working on it?

dhuberdeau:
Hi @qihongl ,  All excellent comments. I added axis labels, legends (where necessary), and titles. I also turned the comments into markup blocks. It does look much better that way!

I think the issue of poor performance when there are a large number of channels stems from the channel width not changing along with the density. This implementation does not giving the user control over the width of channels (e.g. by allowing the user to change the exponent of the sinusoid). It is currently locked at 6, which is reasonable for the methods I've seen it used for. I think in principle one could increase the number of channels to your heart's content, so long as their width was narrowed accordingly. Do you think we should allow users to change the width? It gives more control, but makes it slightly more complicated and offers more opportunity for users to set parameters that are non-optimal or worse.

Thanks! I just did another commit with the changes.

-David

qihongl:
@dhuberdeau Ah I see! Thanks for the explanation. I'm not familiar with what people have been doing on color vision. I guess your judgment is probably more accurate. 

But I thought `n` in the notebook, `iem_example_synthetic_data.ipynb`, is the total number of observations. Maybe I didn't understand the method?

dhuberdeau:
Hi @qihongl, you were right. I had misunderstood your question. 

After playing around with it, it seems like there were some stability issues, especially when the number of observations was increased (for example, for n > 750 it started to break). I found that normalizing the weight matrix helped, but that really it is prone to instability under certain parameter configurations. 

To help solve this, I went ahead and made the exponent a parameter, which will allow users to set the width of the basis functions as well as the number of channels. It seems like playing with those parameters can help eliminate the instability, and might be necessary for the code to be most useful to a general audience, anyway. 

To enable that change (adding the exponent as a parameter), I had to change the way I was defining the channels, so you'll see those changes in there, too.

I hope this fixes it, for the most part. Come to think of it, it might be nice to add a little warning about the possibility for instability. I'll let you take a look at what I've done, but then I will do at least one more commit to document the issue.

Thanks for these great reviews!

-David

qihongl:
@dhuberdeau Thank you very much for being so patient! That's great! I will take a look ASAP!

I think it is useful to know that, for the toy data (with known ground truth) that suppose to be learnable, the algorithm converges to the truth when as long as there is a large number of observations. Otherwise, when it doesn't work in practice, it is hard to understand what's going on.

dhuberdeau:
Hi @qihongl, I think I fixed the issue. See how it works now, and let me know if you find any more unexpected behavior. It was more than instability, I think the key was in the papers, where they said to use rectified basis functions, but then the function weights were allowed to be negative.

Hope that gets us closer. Thanks for all your help!
-David

qihongl:
Hi @dhuberdeau the paper (page 14000) did mentioned that "All channels were normalized to yield an identical maximumresponse of 1." Is this related? I was wondering if the way you normalize the coefficients is grounded on the original paper.

dhuberdeau:
Hey @qihongl, great questions. After reading some other instances of this algorithm being used, and looking at other people's implementations, I don't think there is necessarily a standard way to implement it, aside from following the equations for computing the weight matrix and channel responses. I only chose to normalize after seeing that it fixed the problems we were having with the synthetic data in my example code. 

Although I don't plan to include it for this pull request, I also tested the algorithm with real fMRI data and got the same pattern of results as an implementation that did not normalize. Since the units of channel responses are arbitrary, I'm not sure how much it matters whether it is normalized or not.

So while the code deviates some from the description in Brouwer, 2009, I'd like it to be as general and stable as possible. And along those lines, I also noticed that requiring the data matrix to be "tall" was unnecessarily restrictive, so you'll see in this new commit that I relaxed that requirement. 

Let me know what you think. We could keep it as is, with the normalization, or we could revert back to the more standard/basic implementation and let users normalize themselves if they so desire.

Thanks for your continued help with this.
-David

qihongl:
@dhuberdeau Thanks for the explanation! Right! I also think that a more general implementation is nicer. 

I guess you might be better positioned to decide whether to keep the normalization step. My sense is that if normalization is not used in practice, maybe this step should be optional. On the other hand, I do think it makes sense to show that adding more training example doesn't break the simulation... 

If the code works in practice even without any normalization, but it behaves strangely for the simulation (without normalization, adding more training example hurts), is it possible that the simulation doesn't accurately capture how the algorithm should be used?  So maybe the code is perfectly fine? Let me know if I'm misunderstanding something...

qihongl:
@mihaic The code is probably good, since @dhuberdeau confirmed that the code works for actual data. 

My only concern is that the simulation result is a little strange unless some extra normalization is used. It is possible that this is a sign of a more general issue, but I guess it is probably because the simulation setup is somehow different from the real data. I was asking @dhuberdeau to see what he thinks...

dhuberdeau:
Hi @mihaic and @qihongl, 

I think @qihongl 's concerns were valid; the implementation should not force users to use a slightly less standard method. In any case, for completion, I left the option to do normalization under the hood, in case it becomes useful (which it seems sometimes it might, as in the example data or like in the sub-analysis done in the Heeger, 2009 paper). I'm striving to keep the method valid for as many cases as possible. 

Do you think this is acceptable?  Thank you both for your continued effort!
-David

mihaic:
I think it would be best to answer @qihongl's question before merging.

@dhuberdeau, I know you have been working on this PR for very long. Thank you for your commitment to BrainIAK! If you feel like we should be making more progress, we can think of something new, such as trying to co-opt an extra reviewer. Let us know!

CameronTEllis:
@dhuberdeau any updates on this

dhuberdeau:
> @dhuberdeau any updates on this

Thank you for the reminder. I'm getting back to this and have made progress fixing the example case. I'm cleaning it up now, and should be ready to make another commit attempt in a week or two. -David

qihongl:
Ah, honestly I feel bad about this PR. I guess my request was probably too demanding. I'm sorry about this. 

If I remember correctly, @dhuberdeau was able to replicate the empirical results? I think passing this PR and fix whatever that seems to be a concern in some later PRs is totally reasonable.

vyaivo:
Hi @dhuberdeau and @qihongl -- I'm interested in having some different flavors of IEMs in BrainIAK as well, so I was glad to see this PR. I have played quite a bit with them so I will take a look at the current issues and see if I can help resolve them. I'll update you when I get a chance to run through the code.

dhuberdeau:
@vyaivo That would be so great! I actually have some modifications that I have been meaning to push. It would be better to start off looking at those rather than the latest on Github. I'll do that today, and then you can have at it! Thanks.

vyaivo:
Hi @dhuberdeau,

I took a look at the main `reconstruct/iem.py script` and at `iem_example_synthetic_data.ipynb`.

In the main `iem.py` script, it seems that you are assuming a circular space as the input (e.g. motion direction or orientation). However, the `score` function computes the distance between the predicted & actual stimuli in linear space rather than circular space, which will sometimes give very different results (e.g. 0 degrees and 360 degrees are the same motion direction, so their distance should be 0).

Then in `iem_example_synthetic_data.ipynb`, I think it makes more sense to simulate data such that each voxel has a roughly localized receptive field in feature space. That is the most idealized case. Right now, the code generates responses for each voxel that scale in amplitude and variance for features with larger values. You can see this by plotting a histogram of trial responses to the different stimuli after you generate X:

```
voxel_index = 0
for i in range(0, 3):
    plt.hist(X[i:(i+1)*n_, voxel_index], alpha=0.25)
```

In my tests with these two changes, I did not need to add normalization to get good results. Then, as expected, increasing the number of voxels increased the score (especially when I added more noise to the synthetic data).

I'll clean up my example Python notebook and push it to this PR in the next day or so. I'll also talk to @mihaic about some of my other comments and see if it makes sense for me to review the code, or just make changes once this version merges.


Cheers,
Vy

vyaivo:
Sorry for the delay. I ended up deciding to make several changes all at once, rather than make them piecemeal here. I created a PR on your fork, @dhuberdeau, and wrote a message there explaining the major changes.

The changes I made were for user flexibility, ability to expand the code later, and a couple technical corrections (like scoring predictions in a circular space, as I mentioned above). I wrote my own example Jupyter notebook based on the existing synthetic data notebook that generates data assuming that each voxel is a 1D Gaussian receptive field along the feature space. I also added this data generation code to utils/fmrisim.py.

In addition to modifying the base class and example, I updated test_iem.py to reflect my changes. Let me know if there are any unclear changes. Hopefully this means we are close to merging!

vyaivo:
Hi @qihongl -- I think the issues from your previous review have been resolved, and we should be ready to merge soon. Would you be willing to review this PR again?

qihongl:
Sure! Happy to help with this!

vyaivo:
Hi @qihongl, hope all is going well. I just wanted to check in since it has been about 2 weeks.

qihongl:
Ah sorry, I will do my oral presentation tomorrow. I will work on this asap!

examples/reconstruct/iem_example_illconditioned_X.ipynb
```diff
@@ -0,0 +1,242 @@
+{
```
qihongl: do we still need the old notebooks `iem_example_illconditioned_X.ipynb` and `em_example_synthetic_data.ipynb`?

brainiak/reconstruct/__init__.py
```diff
@@ -0,0 +1,14 @@
+#  Copyright 2018 David Huberdeau & Peter Kok
```
qihongl: and Vy Vo?

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,466 @@
+#  Copyright 2018 David Huberdeau & Peter Kok
```
qihongl: and Vy Vo?

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,466 @@
...
+
+        return score_value
+
+    def get_params(self, deep=True):
+        """Returns model parameters.
+
+        Parameters
+        ----------
+        deep: boolean. default true.
+            if true, returns params of sub-objects
```
qihongl: it seems that `deep` is not used in this func

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,466 @@
...
+        elif self.stimulus_mode == 'halfcircular':
+            domain = self.channel_domain
+            centers = channel_centers
+
+        # define exponentiated function
+        channels = np.asarray([np.cos(np.deg2rad(domain) - cx) **
+                               self.channel_exp
+                               for cx in centers])
+        # half-wave rectification in Brouwer & Heeger
+        # channels[channels < 0] = 0
```
qihongl: do we still need these 2 lines?

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,466 @@
...
+                raise NotImplementedError("This code doesn't currently support"
+                                          " stimuli which are not square "
+                                          "functions in the feature domain, or"
+                                          " stimulus widths that are not even"
+                                          "divisors of the number of points in"
+                                          " the feature domain.")
+
+        C = stimulus_mask @ self.channels_.transpose()
+        # optional normalization
+        # C = C / np.max(C, axis=1)[:, None]
```
qihongl: do we still need these 2 lines?

brainiak/reconstruct/iem.py
```diff
@@ -0,0 +1,466 @@
...
+        Parameters
+        ----------
+        X: numpy matrix of voxel activation data. [observations, voxels]
+            Should contain the beta values for each observation or
+            trial and each voxel of training data.
+        y: numpy array of response variable. [observations]
+            Should contain the feature for each observation in X.
+        """
+        # Check that data matrix is well conditioned:
+        if np.linalg.cond(X) > 9000:
```
qihongl: it might be better to define a variable called `MAX_COND_NUMBER`, since this is re-used later.

qihongl:
I think the update is great. And the notebook examples make a lot of sense. I just had some really minor comments. Thank you very much!

brainiak/reconstruct/__init__.py
vyaivo: I am just going to attribute authorship to myself in the main script. I don't need to claim copyright. Thank you for raising the issue.

vyaivo:
Thanks @qihongl! I addressed your comments and removed another outdated example file (in addition to the ones you flagged). I think we're good to merge.

mihaic:
Thanks a lot for your contributions, @dhuberdeau and @vyaivo! Looking forward to expanding this new `reconstruct` package.

PR #380, opened 2018-09-05
Bug found by user and corrected. Specifically, the gaussian random fi…

CameronTEllis:
Bug found by user and corrected. Specifically, the gaussian random fields had asymmetric variance when the X and Y dimensions were not symmetrical. This led to rippling type pattern in the gaussian fields. Fixed by upsampling the volume into a symmetrical dimensionality and then cropping after generation.


CameronTEllis:
Yes please, do look. Shouldn't be too much. Not sure why Travis is failing but also it won't let me log in to retest the job. Will review after you have been through the PR

lcnature:
Hi @CameronTEllis , can you explain a bit more of the bug? Is it that if the first 3 dimensions are not equal, then the fft of white noise has different number of grids on the 3 dimensions in frequency domain, then the lowest spatial frequencies represented in each dimension are different?

tests/utils/test_fmrisim.py
```diff
@@ -592,6 +592,22 @@ def test_generate_noise():
...
+    mask = np.ones(dimensions)
+    vol = sim._generate_noise_spatial(dimensions, mask)
+
+    assert np.all(vol.shape == dimensions), 'Volume incorrect shape'
+
+    # Check that the variance is similar across dimensions
+    x_var = (np.std(vol, 0) ** 2).mean()
+    y_var = (np.std(vol, 1) ** 2).mean()
+
+    assert abs(x_var - y_var) < y_var * 2, 'Variance is incorrect across dim'
```
lcnature: I kind of feel this criterion is too loose but I also don't know what a good test is.  I would think something like a F-test would be good to test for variance (https://www.itl.nist.gov/div898/handbook/eda/section3/eda359.htm) but you might need more samples (larger grids) to do such test, which makes the computation slow. Anyway, I just feel that testing a ratio of variance would be more appropriate.
An alternative way which may be a bit complex, is to test the degree of smoothness across the 3 dimensions.You may calculate fft of `vol`, and then calculate the indices for each grid in fft(vol) by your `_fftIndgen` function, and calculate the weighted average of indices with the weight being the squared magnitude of the fft, and compare this number with sigma^2. This way, you can check if the spatial smoothness matches the desired one. But of course it still is not checking variance.

brainiak/utils/fmrisim.py
```diff
@@ -1818,6 +1820,13 @@ def _generate_noise_spatial(dimensions,
         logger.warning('4 dimensions have been supplied, only using 3')
         dimensions = dimensions[0:3]
 
+    # If the dimensions are wrong then upsample now
+    if dimensions[0] != dimensions[1]:
```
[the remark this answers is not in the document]
CameronTEllis: The bug only happened when the 1st and 2nd dimension were not the same length but I might as well as add this condition too

CameronTEllis:
@lcnature Yes, that is my understanding of the bug. Then I think that ultimately means the variance is 'truncated' into different ranges

tests/utils/test_fmrisim.py
CameronTEllis: This is a great suggestion. I can use _calc_fwhm for this purpose. I will update the code accordingly.

tests/utils/test_fmrisim.py
```diff
@@ -592,6 +593,92 @@ def test_generate_noise():
...
+
+        # Pull out this coordinate
+        x, y, z = coordinates[i]
+
+        # Is this within the mask?
+        if mask[x, y, z] > 0:
+
+            # Find the the volume sum and squared values
+            v_count += 1
+            v_sum += np.abs(vol[x, y, z])
```
lcnature: If the goal is to calculate the variance, I think `np.abs` is not necessary here.

CameronTEllis:
@lcnature, thank you so much! 
@mihaic, should you do it or I?

mihaic:
@CameronTEllis, you can certainly merge it yourself. Try to set a one-line title please, such as:
`fmrisim: Fix asymmetric variance bug`

PR #379, opened 2018-08-22
added __all__ to utils

szorowi1:
First contribution to brainiak. Helping with #366

PR #382, opened 2018-09-12
reproducing PR 379 in a new branch instead of making PR from master branch

lcnature:
It is recommended to not make PR from one's master branch. This PR reproduces #379 but in a non-master branch.


mihaic:
@lcnature, I think we should credit @szorowi1 for this change and therefore merge PR #379.

As a side note, the Jenkins tests were not running because you were not on the Jenkins admin list. I added all committers to that list now. Also, you can say "Jenkins, add to whitelist." to always test commits from a certain user.

lcnature:
Yes! This duplication PR was just to test out why we had all the failures.

Issue #356, opened 2018-04-06
Replace ndarray with xarray where appropriate


mihaic:
See the xarray docs for "Why xarray?":
http://xarray.pydata.org/en/stable/why-xarray.html

Thoughts?

PR #384, opened 2018-10-04
WIP: Revamped ISC analyses with additional statistical tests

snastase:
I'm re-working and expanding the ISC (and ISFC) code to accommodate several new features. We can now compute either pairwise or leave-one-out ISCs (rather than just leave-one-out). All analyses now also accept 'mean' or 'median' as summary statistics. I've increased the flexibility of inputs to the core ISC functions, and tried to standardize output geometries (time points by voxels). I've moved statistical tests out of the core ISC and ISFC functions—these are now separate functions. In addition to phase randomization, I've added a circular time-shifting test (Kauppi et al., 2014), and group-level permutation and bootstrap tests (Chen et al., 2016). Both the phase randomization and circular time-shift randomization tests operate on the data themselves and call the core ISC function internally to recompute ISCs on the randomized data. Two changes to the phase randomization code: no more randomization across voxels (as this disrupts the spatial autocorrelation of fMRI data), and in the leave-one-out approach, only the left-out subject is phase randomized per iteration. If all subjects are randomized, then N-1 are averaged, the Nth subject will always be correlated with what is effectively a flat time series, yielding an overly tight null distribution and inflated false positive rates (FPRs). Group-level bootstrap and permutation tests operate directly on the ISC values and should better control FPRs (Chen et al., 2016). For pairwise ISCs, bootstrap resampling of subjects and group relabeling operate at the row/column level, not at the level of individual pairs (therefore respecting the correlation structure among pairs). Updated and expanded tests. No statistical tests for ISFC yet, but working on it. I would also recommend changing the module/filename from "isfc.py" to "isc.py", despite the ISC values being a subset of the ISC values, I think ISC is conceptually superordinate and we may want to add things like spatial ISC in the future (which does not fit under the ISFC heading). NB: This is my first major PR, so apologies if I'm missing anything obvious.


buildbot-princeton:

buildbot-princeton:

mihaic:
Jenkins, add to whitelist.

mihaic:
@snastase, please ignore the Travis status for the moment. The failure is unrelated. I'm working on a fix.

qihongl:
Hi, folks @snastase @mihaic sorry about the delayed response! And thank you very much @snastase 
! I'm very excited about this pull request! Though my schedule is pretty crazy recently. If this is not time sensitive, I'd love to help with anything!

manojneuro:
@snastase thanks for working on this. Meir and David Turner are doing some tests on memory utilization of ISC with Searchlight. I'll request them to use this fix as a part of their tests. This will provide an extra validation on many of these updates.

mihaic:
@qihongl, I agree about the geometry comment. See issue #356 for a different idea.

snastase:
@qihongl @mihaic I definitely agree about having a standard input geometry, and any transposes, slicing, etc happening internally. I would advocate for n_samples (typically time points, conditions) by n_features (typically voxels, electrodes, surface vertices), as this seems to be the most widespread convention and makes the most sense for, e.g., brain decoding analyses. I think standardizing this throughout will be critical for overall usability of BrainIAK software. I'm not very familiar with xarrays, but they look very cool. Seems like the ultimate solution would be to generally assume a standardized 2D geometry for neural data (with checks) but also allow for the increased flexibility / explicitness of xarrays.

snastase:
@manojneuro I haven't put a lot of effort into optimizing for speed / computing resources yet—was going for flexibility and readability—so I suspect there are several points where we could speed things up! Current version may run slower than the previous implementation until then, but I haven't thoroughly tested this.

qihongl:
@snastase @mihaic I'm still working on the rest of the code. For now, I just have some minor thoughts.

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+
+    Returns
+    -------
+    iscs : subjects or pairs by voxels ndarray
+        ISC for each subject or pair (or summary statistic) per voxel
+
+    """
+    
+    # Convert list input to 3d and check shapes
+    if type(data) == list:
```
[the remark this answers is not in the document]
qihongl: @mihaic @snastase  Do you think it is feasible to have a common function (perhaps implemented in `util`) that validate and standardize input data?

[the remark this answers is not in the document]
mihaic: @qihongl, we should do that. We already have incipient code in `brainiak.io` and `brainiak.image`.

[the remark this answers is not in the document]
qihongl: @mihaic Awesome!

qihongl:
@snastase Does the current code support statistical testing for ISFC?

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+    for i in np.arange(n_shifts):
+        
+        # Random seed to be deterministically re-randomized at each iteration
+        if isinstance(random_state, np.random.RandomState):
+            prng = random_state
+        else:
+            prng = np.random.RandomState(random_state)
+            
+        # Get randomized phase shifts
+        if data.shape[0] % 2 == 0:
```
qihongl: I think using `n_TRs` instead of `data.shape[0]` is more readable

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+            # Why are we indexing from 1 not zero here? Vector is n_TRs / -1 long?
```
qihongl: is this comment gonna be addressed?

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+        else:
+            pos_freq = np.arange(1, (data.shape[0] - 1) // 2 + 1)
+            neg_freq = np.arange(data.shape[0] - 1, (data.shape[0] - 1) // 2, -1)
+
+        phase_shifts = prng.rand(len(pos_freq), 1, n_subjects) * 2 * np.math.pi
+        
+        # In pairwise approach, apply all shifts then compute pairwise ISCs
+        if pairwise:
+        
+            # Fast Fourier transform along time dimension of data
```
qihongl: Should we pack the following 3 steps: FFT -> shift -> inv FFT as a function, as it is also used in the leave-one-out procedure?

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+        # Compute summary statistic for bootstrap ISCs per voxel
+        # (alternatively could construct distrubtion for all voxels
+        # then compute statistics, but larger memory footprint)
+        if summary_statistic == np.mean:
+            distribution.append(np.tanh(np.nanmean(np.arctanh(isc_sample), axis=0)))
+        elif summary_statistic == np.median:
+            distribution.append(np.nanmedian(isc_sample, axis=0))
+                    
+        # Update random state for next iteration
+        random_state = np.random.RandomState(prng.randint(0, 2**32 - 1))
```
qihongl: This `2**32` is re-used several times. I think this should be a named constant (e.g. MAX_RAND_SEED) in this script.

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+    # Shift bootstrap distribution to 0 for hypothesis test
+    shifted = distribution - observed
+    
+    # Get p-value for actual median from shifted distribution
+    p = ((np.sum(np.abs(shifted) >= np.abs(observed), axis=0) + 1) /
+          float((len(shifted) + 1)))[np.newaxis, :]
+    
+    if return_distribution:
+        return observed, ci, p, distribution
+    elif not return_distribution:
```
qihongl: I kind of like this ... 
```
if P:
    do A
elif not P:  
    do B
```
... structure, compared to ... 
```
if P:
    do A
do B
``` 
since the 1st structure might be more readable for people without programming background. Though I was a little confused initially. I wonder how other people think about this? @mihaic @manojneuro @cbaldassano

qihongl:
I have a discussion point about function organization: 

Currently, ISC with different statistical tests shared some data processing code (e.g. check data shape) at the beginning and data summarization code at the end (e.g. compute p-value, compute summary stats). 

An alternative way of organizing the code is to pack the initial part and the last part as functions. And pack ISC with different testing mechanism into a single master ISC function. And request the testing mechanism as an input argument. 

I guess the current organization is used for very good reasons (the shared processes across different tests are not identical). And I have thought through the all of the caveats. What do you think? Did you had some difficulty merging them? @snastase

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+        # Update random state for next iteration
+        if not exact_permutations:
+            random_state = np.random.RandomState(prng.randint(0, 2**32 - 1))
+        
+    # Convert distribution to numpy array
+    distribution = np.array(distribution)
+    assert distribution.shape == (n_permutations, n_voxels)
+
+    # Get p-value for actual median from shifted distribution
+    p = ((np.sum(np.abs(distribution) >= np.abs(observed), axis=0) + 1) /
```
qihongl: 1. Can we assign a variable name for `float((len(distribution) + 1)))[np.newaxis, :]` to make this line more readable? 
2. This piece of code for p-value computation is used 3 times. Should we pack this as a function?

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+            print("Two-sample exact permutation test using group "
+                  f"randomization with {n_subjects}! "
+                  f"({np.math.factorial(n_subjects)}) "
+                  "iterations.")
+            exact_permutations = list(it.permutations(
+                np.arange(len(group_assignment))))
+            n_permutations = np.math.factorial(n_subjects)
+    
+    # If one group, just get observed summary statistic
+    if n_groups == 1:
```
qihongl: This function doesn't perform input validation for `summary_statistic` at the beginning. So I guess this `if-elif` might fail to catch invalid option for `summary_statistic`?

qihongl:
ah and for line 612: 

`n_groups ` is referred before assignment. Consider replacing `n_groups` as `np.unique(group_assignment)`?

brainiak/isfc.py
mihaic: @qihongl, the two examples you give are not equivalent as written, when `do A` does not necessarily interrupt the control flow (e.g., it is not `return A`).

Also, if the condition tested is the same, we should not repeat it in the `else` clause:
```python
if P:
    do A
else:
    do B
```

I'm not sure I'm completely on topic, but I would say we should have as few returns as possible, generally.

qihongl: @mihaic Right, I agree. 

The main point I want to make here is that, there are about 20 instances of `if P ... elif not P`. I wonder if we should switch all of them to something more standard.

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+    # Manage one-sample and incorrect group assignments
+    elif not group_assignment or len(np.unique(group_assignment)) == 1:
+        n_groups = 1
+        
+        # If pairwise initialize matrix of ones for sign-flipping
+        ones_matrix = np.ones((n_subjects, n_subjects))
+        
+    elif len(np.unique(group_assignment)) > 2:
+        raise ValueError("This test is not valid for more than "
+                         f"2 groups! (got {n_groups})")
```
qihongl: `n_groups` is referred before assignment here. Consider using `len(np.unique(group_assignment))` instead?

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+                  group_labels[1]: group_assignment.count(group_labels[1])}
+
+        # For two-sample pairwise approach set up selector from matrix
+        if pairwise == True:
+            # Sort the group_assignment variable if it came in shuffled
+            # so it's easier to build group assignment matrix
+            sorter = np.array(group_assignment).argsort()
+            unsorter = np.array(group_assignment).argsort().argsort()
+            
+            # Populate a matrix with group assignments
```
qihongl: This line is hard to read/understand. 
Should we add an intermediate step (e.g. by giving `np.full((groups[group_labels[0]], groups[group_labels[1]]), group_labels[0])))` and `np.full((groups[group_labels[0]], groups[group_labels[1]]), np.nan))` some varaible names?

brainiak/isfc.py
mihaic: Indeed. We should not repeat the condition if not necessary. I see no benefit to repeating the condition.

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+    # Infer subjects, groups, voxels and print for user to check
+    n_voxels = iscs.shape[1]
+    print(f"Assuming {n_subjects} subjects, {n_groups} group(s), "
+          f"and {n_voxels} voxel(s) or ROI(s).")
+    
+    # Set up permutation type (exact or Monte Carlo)
+    if n_groups == 1:
+        if n_permutations < 2**n_subjects:
+            print("One-sample approximate permutation test using sign-flipping "
+                  "procedure with Monte Carlo resampling.")
```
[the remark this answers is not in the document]
mihaic: Actually, we do not use `print` at all. We use logging instead, because it is easier to control. This is enforced by `pr-check` via `run-checks`. See the documentation for contributors:
http://brainiak.org/docs/contributing.html#standards

brainiak/isfc.py
```diff
@@ -1,226 +1,1104 @@
...
+    assert distribution.shape == (n_shifts, n_voxels)
+
+    # Get p-value for actual median from shifted distribution
+    p = ((np.sum(np.abs(distribution) >= np.abs(observed), axis=0) + 1) /
+          float((len(distribution) + 1)))[np.newaxis, :]
+    
+    if return_distribution:
+        return observed, p, distribution
+    elif not return_distribution:
+        return observed, p
```
qihongl: If `return_distribution` is `false`, I would still return `distribution` as `None` to make sure the output is always a list of 3 elements. Currently, the user needs to write slightly different code based on the value of `return_distribution`, which is inconvenient.

brainiak/isfc.py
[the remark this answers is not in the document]
snastase: I can add a function to `brainiak.io` or `brainiak.images` that basically performs the reorganization and stacking operation that's currently inside these functions. As far as I can understand, this should have the flexibility to take in either (a) the iterator from e.g. `load_images`, (b) a simple numpy array, or (c) a list of numpy arrays—and standardize this to `n_samples`(TRs, conditions, etc) by `n_features` (voxels, sensors, etc) by `n_subjects` if necessary. I can figure out how to organize the data intuitively for ISFC-related functions, but if we want this to be a general I/O utility, I may need a bit of help making sure the output organization makes sense for other functionalities. Currently, I'm inferring what the user wants from the input they're providing, and then printing the inferred shape so they can make sure the geometry is as intended. Are folks okay with this approach?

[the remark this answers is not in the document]
mihaic: This sounds good.

At this moment, we have a single class method that creates the BrainIAK data format from iterable masked images, `MaskedMultiSubjectData.from_masked_images`. We do not use images with 3D voxels (from `load_images`), because masking is always presents in the pipelines used. The BrainIAK data format is a subclass of `ndarray`, so adding option (b) is easy to do. Option (c) is already supported because lists are iterable. You can add support for even more transformations, but I think we should keep the interface precise by adding new class methods as opposed to accepting multiple input types in the same method; e.g., add `MaskedMultiSubjectData.from_useful_input_type`.

Regarding the shape, as discussed before, I am fine with changing the order of dimensions. The only issue I see is inconsistency within BrainIAK, because we have other methods that take data as features by samples (e.g., SRM, HTFA), although they use `ndarray` directly, not `MaskedMultiSubjectData`.

brainiak/isfc.py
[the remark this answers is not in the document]
snastase: Replacing any print statements with logging functionality...

brainiak/isfc.py
snastase: 1. I think `float()` was my python2 brain thinking about division, but this denominator should equal `n_permutations + 1`.
2. We should—I'll create a separate function for this.
3. Probably doesn't need to be; was just being overzealous about enforcing a particular geometry.

brainiak/isfc.py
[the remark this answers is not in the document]
snastase: For now, I can just change the output shape of MaskedMultiSubjectData.from_masked_images from (n_voxels x n_TRs x n_subjects) to (n_TRs x n_voxels x n_subjects). However, the simplest application of ISC essentially just computes the correlation between two time series; I'd also like the ISC functions to accept, e.g., time series for a single ROI for N subjects. In the context of MaskedMultiSubjectData, this would result in shape (1 x n_TRs x n_subjects); however, I'm not sure this sort of input data really fits the intention of MaskedMultiSubjectData... For example, if I just have a single time series for 10 subjects (e.g., a list of ndarrays), do we expect folks to reorganize everything according to MaskedMultiSubjectData? The only alternatives I can think of are (a) leave some of this functionality in the ISC functions as it's fairly ISC-specific, or (b) write an ISC-specific data (re)organization function that takes care of these cases. Any thoughts? @mihaic @qihongl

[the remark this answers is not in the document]
mihaic: This ROI data sounds generally useful to me. Unless you are fairly certain other analysis methods would not use it, I think we should define it outside of ISC, e.g., `brainiak.image.ROIData`.

brainiak/isfc.py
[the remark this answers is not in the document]
mihaic: Indeed, I think the flag is not necessary. Our documentation points to the guidelines in the official Python documentation. You can use info or debug level for expected operations; here are the options for warnings:
"warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning"
"logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted"
https://docs.python.org/3/howto/logging.html#when-to-use-logging

brainiak/isfc.py
snastase: This is a good idea and do-able, but I might leave it as-is for now... It's a bit tricky because in the pairwise approach, your input data includes N subjects and you need a set of N random shifts to apply to all subjects, while in the leave-one-out approach, you're only applying a single phase shift to the test subject (average of N - 1 subjects is not shifted).

snastase:
@qihongl @mihaic I think I've addressed most of the comments so far. To summarize, I've pulled out separate functions for (1) computing p-values from a null distribution, `brainiak.utils.utils.compute_p_from_null_distribution`, which is somewhat different from the existing `p_from_null`; (2) computing summary statistics, and (3) computing some ISC-specific assumptions about input data—these changes should reduce redundancies in the code. I've also changed `brainiak.image.MaskedMultiSubjectData` to yield a `n_TRs` by `n_voxels` by `n_subjects` array. (That class method seems like a fairly arcane way to basically just `np.dstack` some nibabel images, but I didn't want to make any other major changes.) Other minor changes: (4) statistical tests now mandatorily return the distribution for simplicity's sake; (5) added `MAX_RANDOM_SEED` constant for resampling statistics; (6) replaced `print` statements with `logging` functionality. Thanks @qihongl for all the helpful suggestions!

qihongl:
@snastase Got it! Thank you very much! I'm excited to read the code again! I will try to do another pass this week!

snastase:
@mihaic As far as I can understand, `p_from_null` does something fairly specific. If provided observed values and a null distributions for multiple voxels, it mandatorily evaluates the observed value for each voxel against the max/min values from the null distributions across all voxels. The point of this is to strictly control for multiple tests across voxels, but it's pretty nonstandard. (Furthermore, the two-sided version of the test doesn't account for asymmetric null distributions, see: https://stats.stackexchange.com/questions/25927/doubling-the-tails-in-two-sample-permutation-test). My `compute_p_from_null_distribution` just performs the more basic function of comparing the observed value at each voxel to the null distribution for that voxel. My function also formulates the two-tailed test to account for asymmetric distributions and ensures that unless we're doing an exact test with all possible permutations, we don't get a p-value of zero. I would suggest that my function should supplant `p_from_null`, and that the multiple comparisons component of `p_from_null` be included as a separate function, or a flag in my function call. Maybe @qihongl can correct me if I'm misunderstanding anything in the existing `p_from_null` function.

qihongl:
@snastase thank you very much for addressing all of these comments! I think these are all great enhancements 

@mihaic : I had a conversation with @snastase and @manojneuro . We wonder if we should push this version of ISC/ISFC to brainiak for now and address the remaining issues in separate pull requests, as the current ISC/ISFC code is a qualitative enhancement compared to what's implemented in brainiak currently. Obviously, there are a lot of other issues to think about. For example, the current ISC/ISFC code is using the last axis as the subject axis for the data, and during today's meeting. It seems that we are still considering the possibility of changing that. But issues like this seems to be a brainiak-wide pull request. What do you think?

mihaic:
That makes sense. We should not hold up this PR because of the data format discussion.

qihongl:
@mihaic @snastase Cool! I just made an approval. 
@snastase Do you want to open some pull requests right now to list some pending issues, such as stats testing for ISFC? 
And I was wondering if some processes can parallel processing? The computational demand for ISFC is similar to the FCMA issue. And some stats test, such as permutation testing, seems to be quite parallelizable. 

And thanks again @snastase for this enhancement!

mihaic:
If we change the module name, I think we should keep the previous name as deprecated for a while before removing it. I also created issue #391 to add a deprecation policy, since we don't have one.

mihaic:
@snastase, regarding your f-string commit, I created issue #393 about the minimum version of dependencies we should support.

mihaic:
I think removing (not deprecating) the old module `brainiak.isfc` is actually the way to go considering we are not keeping any function signature unchanged.

In any case, the current approach of duplicating the code is not what we want.

snastase:
Okay, will remove the brainiak.isfc duplicate (and tests) entirely.

I'm a bit confused about how to handle the current test error regarding linked references in the docstring. I'm getting the following message:
```
Warning, treated as error:
/home/travis/build/brainiak/brainiak/brainiak/isc.py:docstring of brainiak.isc.permutation_isc:30:Duplicate explicit target name: "chen2016".
make: *** [html] Error 2
/home/travis/build/brainiak/brainiak
make docs failed
The command "./pr-check.sh" exited with 1.
```
Is this because I have multiple pointers to the same reference? I.e., the Chen2016 reference is used for both bootstrap_isc and permutation_isc.

mihaic:
The error is because of multiple definitions of `chen2016`. You should remove the second of the definitions and replace it with a reference, e.g.:
`The implementation is based on the following publication: [Chen2016]_.`
http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#citations

snastase:
Hey @mihaic, some tests are failing for linux during the conda build, apparently due to problems with pymanopt(?)—any idea what that's about?

mihaic:
It's not related to your PR (I can see the same error on my machine on the master branch). I'm looking into it, but if you're done before I figure it out, we can merge your PR anyway.

snastase:
@mihaic I think all the base functionality @qihongl reviewed is basically ready to go. There are other related topics (e.g., updating the examples)... not sure if those are best handled here or in another PR.

mihaic:
The master branch is supposed to be functional, so I don't think we can postpone updating the examples.

snastase:
@mihaic Okay, the ISC/ISFC example is now functional and updated to use the new code (including new MaskMultiSubjecData shape, etc). I'm not sure why that test would fail. The point was to ensure that the correlation between identical time series was exactly 1. I haven't been able to recreate a situation where the test failed locally. I guess it's possible that the np.corrcoef computation could introduce some tiny deviation and thus break the assertion? I switched the assertion from testing for strictly 1 to np.allclose to introoduce a tiny bit of tolerance.

PR #385, opened 2018-10-10
dev: Pin Python version in Conda build

mihaic:
Conda packages for our dependencies lag behind Python releases. This
causes problems with Travis, e.g.:
https://travis-ci.org/brainiak/brainiak/jobs/435150784

### DIFF
--- a/.conda/bin/build
+++ b/.conda/bin/build
@@ -2,6 +2,12 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+python_version=$1
+if [ -z ${python_version+dummy} ]
+then
+    python_version=3.6
+fi
+
 export BRAINIAK_HOME=$DIR/../../
 
 # See run-tests.sh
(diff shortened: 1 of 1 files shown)


mihaic:
Google hasn't seen the new Travis error anywhere else than in the source code:
https://www.google.com/search?q="dist-info+version+incorrect"

Ideas?

danielsuo:
Weird. Were you able to build locally?

mihaic:
I went back to an older version of Miniconda (https://repo.continuum.io/miniconda/Miniconda3-4.3.30-Linux-x86_64.sh). Same error. Very puzzling.

The error seems egregious, because it compares the version of NumPy to that of BrainIAK.

lcnature:
Hi, a few days ago I might have seen similar error message as in travis when I was probably trying to install brainiak in editting mode on spock. I think when I switched to anaconda 5.1, the errors about dependency on python version went away. Have you tried newer anaconda? Although it is sad that we have to require users to upgrade conda if that is the only way to solve the problem.

mihaic:
Thanks for the tip, @lcnature. We are using the latest version in Travis.

Note that this problem should not affect users, because it only happens when building the Conda package, which the users do not need to do.

danielsuo:
Lol yay

PR #387, opened 2018-10-22
Add an option in BRSA to set the prior of SNR to be "equal" across all voxels

lcnature:


mihaic:
The new errors are unrelated to this PR. I'll fix them.

PR #389, opened 2018-10-24
syntax: Fix Python 3.6 string escape deprecation

mihaic:
From Python 3.6 release notes:
"A backslash-character pair that is not a valid escape sequence now
generates a DeprecationWarning. Although this will eventually become a
SyntaxError, that will not be for several Python releases."
https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

This became a run-checks.sh error and was causing builds to fail in PR #387 starting with pycodestyle 2.4.0:
https://pycodestyle.readthedocs.io/en/latest/developer.html


mihaic:
FYI, this is why the Conda builds are failing on Travis:
https://github.com/matplotlib/matplotlib/issues/12601

As far as I can see on the Matplotlib email list, they will only release a version incorporating the fix next weekend, so I am making the test non-blocking until then.

Issue #400, opened 2019-01-10
fcma: compute_correlation returns zeros for NaNs


snastase:
I just noticed that `brainiak.fcma.util.compute_correlation` returns float zeros when computing the correlation between vectors containing NaNs, which is inconsistent with other correlation computations in Python like `np.corrcoef` and `scipy.stats.pearsonr` (which return `np.nan`). Is this the desired output? Or is it necessary for the high-speed BLAS implementation? Might not be worth bothering with—was just curious.

yidawang:
We designed in this way to prevent the entire computation from crash when encountering the NaN vectors. We can point this out in the docstring if you think this may lead to some potential issue.

snastase:
Okay, makes sense. When invoking `compute_correlation` in the `isfc` function, I may reassign these zeros to NaNs at the output stage to make the output consistent with `isc`. I can add a sentence to the `compute_correlation` docstring mentioning this.

yidawang:
Or can you submit a PR to add an argument to this function, indicating if we will convert NaN to 0? Thanks!

PR #404, opened 2019-01-10
Option to return NaNs in compute_correlation

snastase:
I've added a `return_nans` option to `brainiak.fcma.util.compute_correlation`. Original behavior was to return float 0.0 correlation for any vectors containing NaNs. Now, if `return_nans=False` (the default) we the original behavior, but if `return_nans=True` all zeros are replaced with NaNs before output. This is a very simplistic approach: Another way to do this would be to check the input matrices for NaNs and use this to supply NaNs in the output if `return_nans=True`. For example, if you somehow got a "true" correlation of zero, this would replace the zero with NaN even if there wasn't a NaN in the input. This PR references issue #400.


brainiak/fcma/util.py
```diff
@@ -115,4 +121,9 @@ def compute_correlation(matrix1, matrix2):
                                               0.0,
                                               corr_data,
                                               r2)
+
+    # optionally convert zeros back to NaNs
+    if return_nans:
+        corr_data[corr_data == 0] = np.nan
```
yidawang: It is possible that we have a pair whose correlation is 0. You may want to add this check to line 50, in `_normalize_for_correlation`

snastase: Good point—makes sense. New commit seems to behave as expected.

snastase:
I'm having a hard time understanding the test failures here—seems to be something to do with theano used by test_sssrm...? This is not modified in this PR. Any ideas? @mihaic

mihaic:
See PR #408.

Issue #396, opened 2018-12-17
isc: `phaseshift_isc` overlaps with `brainiak.utils.phase_randomize`


mihaic:
We should not have duplicate functionality.

snastase:
These are not really functional duplicates. I think in theory what we want is (_a_) `phaseshift_isc`, which performs the nonparametric test using phase randomization and returns a null distribution, _p_-values, etc; and (_b_) a more generic `phase_randomize` utility that simply takes in time-series and phase-randomizes them. I think having this `phase_randomize` utility is useful even outside the context of ISC analysis. And ideally, the `phaseshift_isc` function should use the `phase_randomize` function internally. This was a little tricky because the pairwise and leave-one-out ISC approaches apply the phase randomization slightly differently, but should be work now—see #405 (commit [dba670a](https://github.com/brainiak/brainiak/pull/405/commits/dba670a9686841a4232d9867ec440488a20b05eb). I've updated `phaseshift_isc` in `isc`, `utils` and tests to match.

mihaic:
There's a way to close issues with commit messages such that it is clear they were fixed. I think we should use it in the future:
https://help.github.com/articles/closing-issues-using-keywords/

Issue #399, opened 2019-01-09
isc: output dimensions for ISFC analysis


snastase:
I want to solicit opinions about the ideal output dimensions for `brainiak.isc.isfc`; just had a conversation with @manojneuro offline about this. Currently, if we input an array with shape `(n_TRs, n_voxels, n_subjects)`, the vanilla `isc` function returns an `(n_subjects, n_voxels)` (or `(n_pairs, n_voxels)`) array of ISC values—this makes sense to me based on our discussions about standardizing dimension order in BrainIAK, because these ISC values are now our samples of interest per voxel. The downstream statistical tests for ISCs are expecting this `n_subjects` by `n_voxels` structure—typically your test will return a single statistic that collapse over the first dimension, leaving you with a single value (e.g., a p-value) per voxel. In contrast, if we supply the same array to `isfc`, we get an output array of ISFCs with shape `(n_voxels, n_voxels, n_subjects)`. This sort of makes sense in a vacuum, but doesn't really match the logic of the `isc` function, and also won't play nicely with statistical tests without some manual re-structuring. I think my favorite option here would be to use `squareform` to vectorize the `n_voxels` by `n_voxels` matrices per subject into something like `n_connections` and put this vector in the second dimension, so the output would be shaped `(n_subjects, n_connections)`. With 20 subjects and 1000 voxels, this will yield an `isfc` output shaped `(20, 499500)` (where `1000 * (1000 - 1) / 2 = 499500`). This would play nicely with the `compute_summary_statistic`, `bootstrap_isc`, and `permutation_isc` tests without any external reshaping by the user, yielding an output vector with 499500 values (e.g., mean ISFCs, p-values, etc), which can then be supplied to `squareform` to recover the `(1000, 1000)` shape. This might be a little less intuitive at first, but seems more internally consistent and efficient. Any thoughts on this? @qihongl @mihaic

mihaic:
Sounds reasonable. In terms of API stability, considering we have not released a new BrainIAK version since PR #384, I see no issue making the change.

Issue #397, opened 2018-12-17
utils: Document difference between compute_p_from_null_distribution and p_from_null


mihaic:
See:
https://github.com/brainiak/brainiak/pull/384#issuecomment-438000981

snastase:
As far as I can tell `p_from_null` was performing a specific function that was only used in the old `isfc` function, and is no longer referenced anywhere in `brainiak`. In this case, I think `compute_p_from_null_distribution` supersedes `p_from_null`, and `p_from_null` can be removed. However, I may shorten the name of `compute_p_from_null_distribution` as I only named it this to not overlap with the pre-existing `p_from_null`.

mihaic:
That makes sense. Please remember to include a news item about the removal of `p_from_null` (see point 7 http://brainiak.org/docs/contributing.html#how-to-contribute). Looking forward to the PR!

snastase:
Replaced old `p_from_null` with newer functionality from `compute_p_from_null_distribution` in #405 (commit [daa5d41](https://github.com/brainiak/brainiak/pull/405/commits/daa5d4190db01ebd7d4340774705b1bb6fb5eca6)). Note that the new functionality now uses the older simpler function name. This also required removing the `ecdf` functionality which is no longer used anywhere in BrainIAK (commit [aa0c37d](https://github.com/brainiak/brainiak/pull/405/commits/aa0c37d11bc5b3602caee33dde6eb45755177c83).

PR #405, opened 2019-01-12
Handling NaNs in ISC/ISFC (and ISFC output shape)

snastase:
I've added a `tolerate_nans` option to the `isc` and `isfc` functions to better handle NaNs, particularly when averaging time series for *N*–1 subjects in the leave-one-out approach. NaNs are typically introduced prior to ISC analyses by z-scoring voxels with zero variance (due to susceptibility, limited EPI FoV; see issue #398). The default value is `True`, which accommodates (i.e., ignores) any NaNs when computing the average of *N*–1 subject time series by using `np.nanmean`. If the user wants to handle NaNs beforehand, they can set `tolerate_nans=False`, which will return a NaN even if there's a single NaN in the time series for any single subject—but this uses `np.mean` rather than `np.nanmean`, which I believe is faster. Alternatively, the user can supply a float between 0.0 and 1.0 indicating the threshold proportion of subjects with non-NaN values needed to keep the voxel. For example, if the threshold is 0.8, any voxel with >= 80% subjects with non-NaN values will be kept and the average of *N*–1 subject time series will be computed to tolerate the remaining NaNs; if a voxel has < 80% subjects with non-NaN values (i.e., > 20% of subjects have NaNs), the voxel is excluded from the computation. In all cases, any voxel with a NaN in the time series across all subjects is removed prior to any computation, as this will always result in NaN in the final output (to reduce compute time). If voxels with NaNs are censored out, the resulting output will still match the original number of voxels with NaN values padded back in appropriately. I'm arguing to leave the default value as `True` rather than some particular proportion, because I'm not sure any specific proportion is actually reported consistently in the literature. I had to add an option to return NaNs in `brainiak.fcma.util._normalize_for_correlation` to make sure the `isfc` function (which uses `fcma.util.compute_correlation` for speed) returns NaNs consistent with `isc` (and other Python correlation functions; see issue #400). This PR relies on a merge of PR #404. Finally, I added a `vectorize_isfcs` option to `isfc`, which by default (`True`) vectorizes the upper triangle of the ISFC matrices, yielding a 2D `n_subjects` (or `n_pairs`) by `n_connections` (pairs of voxels) array; if `vectorize_isfcs=False`, we keep the square (redundant) ISFC matrix and output is shaped `n_subjects` (or `n_pairs`) by `n_voxels` by `n_voxels`. This is more consistent with the shape of the `isc` output, and is more compatible with the subsequent statistical tests.


snastase:
@mihaic Any idea what's going on with this gcc mpich dependency error in the Travis Xcode 7.3 Mac build?

snastase:
@qihongl @manojneuro Can I get another pair of eyes on this? Thanks!

qihongl:
@mihaic @snastase I'd love to take a look on this! Sorry about the delayed response. I'm a little over overwhelmed recently XD

brainiak/isc.py
```diff
@@ -411,6 +490,126 @@ def compute_summary_statistic(iscs, summary_statistic='mean', axis=None):
...
+        nans += ~(np.sum(~np.any(np.isnan(data), axis=0), axis=1) >=
+                  data.shape[-1] * tolerate_nans)
+        logger.info("ISC computation will tolerate voxels with at least "
+                    "{0} non-NaN values: {1} voxels do not meet "
+                    "threshold".format(tolerate_nans,
+                                       np.sum(nans)))
+
+    else:
+        logger.info("ISC computation will not tolerate NaNs when averaging")
+
```
qihongl: I wonder if it is better to explicitly do `if tolerate_nans is False:` here, and return something like  "invalid input" for the `else:` block (e.g. if `tolerate_nans` is a string)?

snastase:
Newest commits should resolve issues #396 and #397.

mihaic:
@snastase, we need to review and merge PR #414 for the Jenkins Linux test to pass.

snastase:
Hey @mihaic—do you know what's breaking the Linux build? I thought I had successfully merged in the #414 fix at [57a9d09](https://github.com/brainiak/brainiak/pull/405/commits/57a9d09ea71c1bd6b5dc5b8671d997903e5fdabf). Any ideas?

mihaic:
@snastase, no sure. I'll try again.

mihaic:
Jenkins, retest this please.

tests/utils/test_utils.py
```diff
@@ -352,6 +368,13 @@ def test_check_timeseries_input():
     assert n_voxels == 1
     assert n_subjects == 10
 
+    # Check that 4-dimensional input array throws error
+    array_4d = array_3d[..., np.newaxis]
+    try:
+        (data_array_4d, _, _, _) = _check_timeseries_input(array_4d)
+    except ValueError:
+        pass
```
mihaic: Pytest has a nice way to check exceptions:
https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

Up to you.

snastase: Cool! Incorporated this and looks a bit cleaner.

mihaic:
Not sure what's up with Jenkins. Please feel free to ask it to test again whenever the error doesn't make sense. (See my comment from yesterday for the exact phrasing.)

snastase:
Jenkins, retest this please.

mihaic:
I hope you're not waiting for me for this PR. I could go through it again to leave more code-related comments, but I think it is more important to reach an agreement on the method changes and for that we should hear from @qihongl.

snastase:
@qihongl and I had some discussions about this offline as well. I wanted to summarize the changes here because this sort of sprawled beyond the original scope of the PR. First, we can now accommodate NaNs when averaging _N_–1 time series in the leave-one-out approach, and can support an arbitrary proportion of NaNs to accommodate (masking out any voxels that exceed this threshold; see #398, #400, #404). Second, and importantly, the output of (symmetric) ISFC analysis can be represented as either a 3-dimensional array of redundant square matrices with shape `n_subjects` (or `n_pairs`) by `n_voxels` by `n_voxels` or a tuple containing the condensed triangles of the ISFC matrices and the diagonals (see #399). The goal of the latter arrangement is to keep the ISFC outputs 2-dimensional so they can easily be passed into statistical tests. We also provide a `squareform_isfc` function to easily switch between the two (redundant square and condensed vector) representations. Third, the `isfc` function can now take a separate `targets` array as input (see #409). In this approach, the ISFC computation is not _among_ voxels in the standard input data, but between each voxel in the input and each voxel (or ROI, or whatever) in the `targets` array, which could, for example, be a separate set of target ROIs. This typically results in non-square, asymmetric ISFC matrices which are output in the square (not condensed) format, and is only implemented for the leave-one-out approach. Finally, we remove/rearrange some duplicate functionality, namely `phase_randomize` and `p_from_null` in `utils` (see #396, #397). All of these modifications have defaults that shouldn't disrupt preexisting usage; except the ISFC output shape. Currently, my thought is to return the ISFCs in the condensed format, but I could be convinced otherwise. I can update any issues this might introduce in tutorials.

qihongl:
Awesome! I will read the new code this week!

qihongl:
I think all of these changes look great!

snastase:
@mihaic Is this reasonable for news? I also ran through @manojneuro's tutorials and make a couple minor fixes to ensure that rolling this out won't break things.

mihaic:
It looks great! I'll merge it as soon as testing is done.

mihaic:
Jenkins, retest this please.

Issue #406, opened 2019-01-13
SRM doesn't apply mu in transform() 


qihongl:
I recently did a simulation and realized that brainiak SRM can't align two trajectories that different by a vector shift, though it should be able to align them by estimating the correct `mu`. 

Here's the <a href="https://github.com/qihongl/scratch-srm/blob/master/srm_demo.ipynb">notebook</a> containing the simulation. By uncommenting some code block, the notebook can show that 
i) the shifting can be fixed by doing mean centering 1st 
ii) SRM can fix rotational difference (i.e. trajectory 2 = some orthogonal transformation of trajectory 1), as expected

@cameronphchen thinks there is a bug. Here's what he said: 
"
I think this might be due to this line https://github.com/brainiak/brainiak/blob/efce8ae9e46333d2e43d38ea0afff0f6bba12fff/brainiak/funcalign/srm.py#L263
In the implementation, because fMRI data is usually preprocessed by z-scoring, seems like the re-centering part wasn't properly taken care of. If you change the above line to be W^T(X-mu), you should be able to get the expected results. 
" 

We tried it, which indeed fixed the issue. I can make a PR later, but thought I would make a note first.

PR #407, opened 2019-01-14
srm: subtract the subject-specific mean in .transform()

qihongl:
This is a fix described in #406 . With @cameronphchen 's fix, SRM can align two trajectories that differ by a vector shift by using the estimated `mu`.

### DIFF
--- a/brainiak/funcalign/srm.py
+++ b/brainiak/funcalign/srm.py
@@ -260,7 +260,8 @@ class SRM(BaseEstimator, TransformerMixin):
         s = [None] * len(X)
         for subject in range(len(X)):
             if X[subject] is not None:
-                s[subject] = self.w_[subject].T.dot(X[subject])
+                s[subject] = self.w_[subject].T.dot(
+                    X[subject] - self.mu_[subject][:, np.newaxis])
 
         return s
 


qihongl:
Sorry, I didn't expected this to trigger some run test failures and I will take a look later...

qihongl:
Based on the travis report, the failure came from the tests for `sssrm`. It seems that `sssrm` is using (and only using) the `_init_w_transforms` function from `srm`, but the error occured when calling the `fit` function of `sssrm`. So I'm a little confused about why changing the `transform` function from `srm` would break that... Any clue?

mihaic:
See PR #408.

qihongl:
@TuKo Ah I see. `transform_subject()` is solving a standard procrustes problem. Is there a direct extension for procrustes with an intercept? Or does it make sense just do some mean centering, align by standard procrustes, and add the mean back? I'm frankly now sure what's the best thing to do here. 

Right, the documentation should be updated. And later on I can modify my notebook to show detSRM doesn't align two trajectories differ by a vector shift whereas SRM does, and upload it to `example/`.

qihongl:
@TuKo I had a conversation with Sam and I guess for the transform_subject() function with intercept (i.e. `mu`). We probably want to do some mean centering for the estimated S and X_new, then align, and add the mean of S back. In this case, I was wondering if the user should do this manually? Does it seem strange to do this for the user implicitly?

TuKo:
@qihongl `transform_subject()` should compute the mean and substract it from data to compute the new subject's transform (see [here](https://github.com/brainiak/brainiak/blob/c590b7d1939b7ce4c38929dac1af78845a0ef13c/brainiak/funcalign/srm.py#L305) ). BTW, the `mu` obtained for the new subject should be returned from the function as well.

Please note that DetSRM, RSRM and SS-SRM should follow SRM by removing mean. 
I was suggesting to document the other classes, but you are welcome to modify them too. :)

qihongl:
I see! Thanks! @TuKo

mihaic:
@qihongl, any progress?

mihaic:
@qihongl, are you planning to finish this PR?

qihongl:
Ah my github is not linked to my main email address so I didn't notice this. And I totally forgot this PR. 

Right I should finish this PR -- I will work on it get back to you guys in 2 weeks!

qihongl:
Sorry I finally got to this... Can I make a new clone of the latest brainiak and start this from scratch?

mihaic:
If that works better for you, go ahead, @qihongl.

qihongl:
sounds great!

qihongl:
@mihaic 
I want to update the documentation for DetSRM, RSRM and SS-SRM to mention that they don't apply mean removal as SRM. Should I just add this comment in the class doc string? Thanks!

mihaic:
Yes, the class docstring is a good spot. Of course, it would be even better to change the other code as well. Up to you.

PR #417, opened 2019-04-08
docs: Move PR #384 NEWS entry to newsfragments

mihaic:
This will add it to the next release, not v0.8, as it should be.
In PR #405, I mistakenly suggested editing NEWS directly.


mihaic:
Edited to say this corrects PR #405.

PR #418, opened 2019-04-08
dev: Drop support for Python 3.4, add 3.6 testing

mihaic:
Mypy dropped support for Python 3.4 in version 0.700. Technically, we
could separate the Python version in developer requirements from the
Python version in user requirements (where Mypy is not a dependency),
but I think that would complicate our setup too much.

Also switch to Ubuntu 16.04 for Travis testing, since support for 14.04
ends in April 2019.

PR #419, opened 2019-04-08
Added matrix multiplication to ISC

CameronTEllis:
Added matrix multiplication to ISC, rather than the looping through voxels. From testing it provides between x5 and x10 speed up. No changes were made to the pairwise analysis because I wasn't sure how to deal with squareform. A test was also added for the newly added _corr_mat function

### DIFF
--- a/brainiak/isc.py
+++ b/brainiak/isc.py
@@ -50,7 +50,6 @@ The implementation is based on the work in [Hasson2004]_, [Kauppi2014]_,
 import numpy as np
 import logging
 from scipy.spatial.distance import squareform
-from scipy.stats import pearsonr
 import itertools as it
 from brainiak.fcma.util import compute_correlation
 from brainiak.utils.utils import (phase_randomize, p_from_null,
@@ -166,14 +165,14 @@ def isc(data, pairwise=False, summary_statistic=None, tolerate_nans=True):
     elif not pairwise:
 
         # Preset array
-        iscs_stack = np.zeros((data.shape[2], data.shape[1])) 
+        iscs_stack = np.zeros((data.shape[2], data.shape[1]))
 
-        # Leave one out loop    
+        # Leave one out loop
         for s in range(data.shape[2]):
 
             # Get the individual and the average of the others
-            individual = data[:, :, s].T 
-            other = np.mean(np.delete(data, s, axis=2), axis=2).T
+            individual = data[:, :, s].T
+            other = mean(np.delete(data, s, axis=2), axis=2).T
 
             # Perform the row-wise correlation
             iscs_stack[s, :] = _corr_mat(individual, other)
@@ -469,11 +468,11 @@ def _check_targets_input(targets, data):
 
 def _corr_mat(mat_1, mat_2):
     """Computes matrix-wise correlation values for ISC
-    
-    This takes in as input two matrices of voxel by time and returns 
-    a correlation value for each voxel (ISC values). This is an efficient 
+
+    This takes in as input two matrices of voxel by time and returns
+    a correlation value for each voxel (ISC values). This is an efficient
     computation to perform correlations across many voxels
-    
+
     Parameters
     ----------
     mat_1 : 2D array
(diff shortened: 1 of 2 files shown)


mihaic:
After fixing current errors, testing will still fail until PR #418 is merged to master and here.

snastase:
@CameronTEllis Awesome, great. I think I can hack the `squareform` in the pairwise approach using `itertools.combinations`—will time it and get back to you.

snastase:
Hey @CameronTEllis, do you have any code snippet showcasing the speed-up with this method? Is there some C-related configuration I need to set up or something? My cursory comparison of the new and old implementations on a vaguely realistically-sized input shape `(300, 30000, 30)` didn't give me a speed-up. I would _expect_ a speed-up, so I'm probably just doing something dumb.

Screen Shot 2019-04-09 at 6 27 08 PM

CameronTEllis:
@snastase I am not sure why you are getting that behavior. When I do the same test I get different results. 

image

However I was actually underwhelmed with what I got. When testing with smaller numbers I did much better:
image

It turns out that the number of time points really slows down this new analysis:

image

snastase:
Hey @CameronTEllis, sorry for the delay—I tried to dig a bit deeper into this. I'm going to split this into a couple comments. First of all, I made some adjustments to your `_corr_mat` function so as to only compute the de-meaned variables once (slight speed-up) and make it a bit more flexible (supply axis argument to compute column- or row-wise). This updated version that I'm calling `array_correlation` might be generally useful, so I'm currently putting it in `utils` and importing into `isc`. My tests suggest that this is quicker than existing functions like `pearsonr`.
Screen Shot 2019-04-17 at 6 58 36 PM

snastase:
I was also wondering if the method for computing the mean over _N_–1 subjects was slowing things down. So first I tried a couple different ways of getting the _N_–1 array... but the original way using `np.delete` was fine:
Screen Shot 2019-04-17 at 12 59 33 PM

Also checked to see if manipulating lists of subjects might be faster, but no:
Screen Shot 2019-04-17 at 1 30 11 PM

Just wanted to look into possible variants here based on the code from other people in the lab.

snastase:
Next, I moved into the `isc` function. First of all, I tried the special case of _N_ = 2 subjects with the old voxel-loop code and the new `array_correlation` code:
```
def old_two(data):
    voxel_iscs = []
    for v in np.arange(data.shape[1]):
        voxel_data = data[:, v, :].T
        if n_subjects == 2:
            iscs = pearsonr(voxel_data[0, :], voxel_data[1, :])[0]
        voxel_iscs.append(iscs)
    iscs_stack = np.column_stack(voxel_iscs)
    return iscs_stack

def new_two(data):
    iscs_stack = array_correlation(data[..., 0],
                                   data[..., 1])[np.newaxis, :]
    return iscs_stack
```
And the new `array_correlation` gives us a significant speed-boost.
Screen Shot 2019-04-17 at 7 14 03 PM

snastase:
Interestingly, for pairwise ISCs, I can't figure out how to get a useful speed boost over the original code. In the original code we loop over voxels, then use `np.corrcoef` to compute correlations across every pair of subjects. In the new approach, we would loop through pairs of subjects and compute correlations across all voxels at once using `array_correlation`. In my testing, the latter approach was only faster with very small numbers of subject (e.g., `n_subjects <= 5`). With more typical data, the voxel loop with `np.corrcoef` was *dramatically* faster. Open to any suggestions here.
```python
    def old_pairwise(data):
        data = np.swapaxes(data, 2, 0)
        voxel_iscs = []
        for v in np.arange(data.shape[1]):
            voxel_data = data[:, v, :]
            iscs = squareform(np.corrcoef(voxel_data), checks=False)
            voxel_iscs.append(iscs)
        iscs_stack = np.column_stack(voxel_iscs)
        return iscs_stack
        
    def new_pairwise(data):
        isc_pairs = []
        for pair in combinations(np.arange(data.shape[-1]), 2):
            isc_pair = array_correlation(data[..., pair[0]],
                                         data[..., pair[1]])
            isc_pairs.append(isc_pair)
        iscs_stack = np.array(isc_pairs)
        return iscs_stack
```
Screen Shot 2019-04-17 at 9 38 18 PM

snastase:
Finally, we get a bit of a speed-up in the leave-one-out approach. Part of this speed-up comes from using `array_correlation`, but I think part of it also comes from your approach of using the subject index `s` as a view into the `data` array when creating the left-out subject and the mean of _N_–1 subjects—rather than looping through the actual `data` array and pulling out each `subject`.
```python
def old_loo(data):
    data = np.swapaxes(data, 2, 0)
    voxel_iscs = []
    for v in np.arange(data.shape[1]):
        voxel_data = data[:, v, :]
        iscs = np.array([pearsonr(subject,
                      np.mean(np.delete(voxel_data,
                                     s, axis=0),
                           axis=0))[0]
             for s, subject in enumerate(voxel_data)])
        voxel_iscs.append(iscs)
    iscs_stack = np.column_stack(voxel_iscs)
    return iscs_stack

def new_loo(data):
    iscs_stack = []        
    for s in np.arange(data.shape[2]):
        iscs_stack.append(array_correlation(
            data[..., s],
            np.mean(np.delete(data, s, axis=2), axis=2)))
    return np.array(iscs_stack)
```
Screen Shot 2019-04-17 at 10 01 33 PM

Note that I tested all this on my puny laptop with a zillion Chrome tabs open, so take the absolute timings with a grain of salt.

CameronTEllis:
@snastase This is awesome, thorough work that could be a huge speed up (5x in the most common use of ISC). Making `array_correlation` a utils function is a good idea. In general, I think that if this method speeds up most use cases and shows little speed difference in some cases then that is a green light to add it in. If you agree then I will review your code and we can get this in

snastase:
Closing this due to changes merged via #420.

PR #420, opened 2019-04-18
Updated ISCs with array-style correlation

snastase:
@CameronTEllis I moved your new function to `brainiak.utils.utils.py` and modified it a bit (renamed to `array_correlation`). Expanded the tests a bit as well. For the `isc` function, we're now using this for the two-subject case and for the leave-one-out approach for substantive speed boosts. I've left the pairwise approach largely as it was (using a voxel loop and `np.corrcoef`, as the `array_correlation` version is only faster with relatively few subjects (due to large number of subject pairs). The `run-checks.py`, `test_utils.py`, and `test_isc.py` all pass on my end.


snastase:
@mihaic Oops, I was not expecting this to show up separately here... Was pushing to @CameronTEllis's branch, but expected this to show up in the existing PR #419. Should I close this PR?

mihaic:
@snastase, this PR is from your fork. If you can see all the commits you want in PR #419, close this PR.

CameronTEllis:
I think we should keep this PR and close mine (419) since that PR doesn't have those edits. If @mihaic agrees I will review this PR now

mihaic:
You could also add the commits to the other PR. Just push to the PR branch. Up to you.

I'm happy to help where I can, but I think not much around here really requires my approval.

Issue #416, opened 2019-04-04
SRM could be faster and more memory efficient


hugorichard:
Current implementation of SRM could be enhanced when the number of subjects and sessions are big. 
Our experiments show that our implementation is more memory-efficient and faster while having similar accuracy (measured with test R2 reconstruction error) than ProbSRM .
I have a PR ready for this. If you think it is interesting please let me know.
Best,
Hugo RICHARD

mihaic:
Absolutely! Please submit a PR.

qihongl:
Thanks @hugorichard ! This is a really exciting addition! 

Do you have some simulation results showing it is qualitatively similar to the old SRM and faster when the data is big enough? I would love to take a look. Besides qualitatively verifying the code, I think these results will be useful to guide users' decision. 

I'm not sure I'm the best person to review the technical stuff, such as the detail of EM updates. Can we invite some reviewers who actually work on SRM?

hugorichard:
Yes you are right the example does not show that FastSRM is a better choice than the old SRM. It just shows that the implementation is running and how to use the API. I tried to use the same example as in the old SRM but I cannot get my hands on the masker that was used to mask the movie watching data which prevents me from using an interesting atlas to perform the reduction. The mask given in  brainiak/examples/funcalign/download-data.sh ("data/MNI152_T1_3mm_brain_mask.nii") does not match the shape of the data given in the same file ("data/movie_data.mat").

Let me share with you some results on a real dataset (Forrest data using the full brain, 19 subjects and 7 sessions). The data may be too big for a small example but it shows why FastSRM should be used instead of FastSRM.
[memory_usage_forrest_all.pdf](https://github.com/brainiak/brainiak/files/3223971/memory_usage_forrest_all.pdf)
[fit_time_forrest_all.pdf](https://github.com/brainiak/brainiak/files/3223972/fit_time_forrest_all.pdf)
[perfs_forrest_basc_n_iter_10.pdf](https://github.com/brainiak/brainiak/files/3223980/perfs_forrest_basc_n_iter_10.pdf)


The experiment is done as follows:
All sessions but one are used to compute spatial maps for every subjects. Then spatial maps and data from the left-out session of all subjects but one are used to compute the shared response in the left-out session. At last, the shared response during the left-out session and  the spatial maps of the test subject are used to predict the data of the left-out subject during the left-out session. The performance of the model is measured by comparing the prediction and true data using the coefficient of determination (mean r2 score).

[...]

PR #421, opened 2019-04-29
Fastsrm

hugorichard:
Solves #416 .
This PR introduces FastSRM.  On large datasets, it is faster and more memory efficient that the original algorithm with equivalent performance.

The API used is a bit different from the original SRM algorithm for the following reasons:
- This algorithm is intended to be used on very large dataset that do not hold in memory. Hence input of the algorithms are arrays of .npy files (masked data) of shape [n_subjects, n_sessions] and not a list of np arrays.
- The .npy files are arrays of shape (n_timeframes, n_voxels) and not (n_voxels, n_timeframes). This is to be consistent with the sklearn API (n_samples, n_features).


buildbot-princeton:

buildbot-princeton:

mihaic:
Jenkins, add to whitelist.

mihaic:
@hugorichard, thank you very much for the PR.

Hopefully the SRM authors and contributors will be able to review it soon. @TuKo, @hejiaz, @cameronphchen?

hugorichard:
The tests failed because there is the joblib package missing (which I use for multi-processing). 
However it is in the requirement-dev.txt . It works when i run ./pr-check on my end. Is this normal ? What should I do to fix this ?
Hugo

mihaic:
If you are adding it as a dependency of BrainIAK (not just for testing), you must add it to `install_requires` in `setup.py`. You should then remove it from `requirement-dev.txt`, because we do not duplicate BrainIAK dependencies there.

hugorichard:
Thanks ! I corrected this and it still does not run. It seems to be related to MPI although I do not make use of it: 
```
from mpi4py import MPI
E   ImportError: dlopen(/Users/travis/miniconda3/conda-bld/brainiak_1556786193685/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.6/site-packages/mpi4py/MPI.cpython-36m-darwin.so, 2): Library not loaded: @rpath/libmpi.1.dylib
```

mihaic:
I'm looking into the error. Please ignore it for the moment, since it only affects the Conda build.

Note that `typing` should not be a dependency, because it is included in the standard library starting with Python 3.5, which is the minimum required version for BrainIAK.

mihaic:
Sorry for the last remark. I saw `typing` in the diff and though you added it.

mihaic:
@qihongl, I hear you are also knowledgeable about SRM. Review maybe?

manojneuro:
@hugorichard We were busy with the tutorials (https://brainiak.org/tutorials) release, but @qihongl will review this soon.

hugorichard:
@manojneuro Thank you for your answer ! Don't hesitate to contact me if you need more information.

hejiaz:
@hugorichard Thank you for your pull request. I'm working on SRM development, so I could try to review this pull request. Do you have any paper/write-up about this method? A paper about it would be really helpful when reviewing this request. Thanks!

hugorichard:
I have just uploaded a paper on Arxiv: https://arxiv.org/pdf/1909.12537.pdf

brainiak/funcalign/fastsrm.py
```diff
@@ -0,0 +1,1014 @@
...
+        if self.verbose is True:
+            logger.info("[FastSRM.fit] Finds shared "
+                        "response using reduced data")
+
+        shared_response_list = fast_srm(
+            reduced_data,
+            n_iter=self.n_iter,
+            n_components=self.n_components,
+        )
+
```
hejiaz: This fast_srm function is basically DetSRM. Since this part does not use parallelization anyway, could you re-use the functions (DetSRM and SRM) in srm.py to do this fitting?

brainiak/funcalign/fastsrm.py
```diff
@@ -0,0 +1,1014 @@
...
+            # Remove files in temp folder
+            paths = glob.glob(os.path.join(self.temp_dir, "*.npy"))
+            for path in paths:
+                os.remove(path)
+
+            self.low_ram = low_ram
+
+    def fit(self, imgs):
+        """Computes basis across subjects from input imgs
```
hejiaz: Optional: You could consider providing an API option that is the same as SRM, where the input is a list of arrays of size voxel-by-time. Also, the input data of FastSRM has a different orientation compared with all the other algorithms in the funcalign module. It might be better to keep everything consistent.

hugorichard:
@hejiaz Thanks a lot for your comments (and for reviewing this PR).

brainiak/funcalign/fastsrm.py
hugorichard: fast_srm does have one thing that DetSRM does not have which is the "low_ram" mode. In some cases even the reduced dataset might not hold in memory (when the number of sessions is very big and the number of regions in the atlas are big) in which case it can be useful to write reduced data on disk. 
Typically this "low_ram" mode is useful when the number of voxels is a lot less than the number of sessions x number of subjects x number of reduced components.
This functionality is not mentionned in the paper and I did not use it in the experiments. I could therefore get read of it but as datasets will grow bigger it might be useful at some point.
Another option would be to call DetSRM if low_ram mode is not used and to use the current implementation if low_ram mode is used.

brainiak/funcalign/fastsrm.py
hugorichard: I'll add this option. However being able to provide a list of .npy file instead of a list of arrays allows FastSRM to handle datasets that are difficult to hold in memory so I would like to keep this possibility.

brainiak/funcalign/fastsrm.py
hejiaz: I think it is better to call DetSRM/SRM if low_ram is not used and to use the current implementation if low_ram mode is used. An advantage of this is that when low_ram is not used, you can also apply ProbSRM (or any other algorithms similar to SRM).

brainiak/funcalign/fastsrm.py
hejiaz: Yes, keeping both options would be optimal.

hugorichard:
(Although I have pushed some commits it still needs some work before a second pass can be done I'll tell you when I am done)

hugorichard:
Now I have pushed a stable and fully tested version. I still need to update the example so that it complies with the new API. I did a bit more than requested so it was a lot of work but now some new features are available:
- temporary files can (and should !) be cleaned when the fitted model is not used anymore by calling the .clean method
- various input format are accepted output shapes are coherent with input shapes
- aggregate attribute (shared responses are averaged if aggregate is set to "mean" (default) otherwise a shared response per subject is returned like in DetSRM or ProbSRM)
- a method add_subjects is similar to transform_subject in DetSRM. the difference is that the additional subject is directly add to internal basis_list so that everything behaves like if this subject was present when fitting
- when no atlas is provided FastSRM defaults to DetSRM however low ram mode is still available so that this provides a way to apply vanillia DetSRM even if the data is difficult to hold in memory.

hugorichard:
Update: I don't get why test don't pass online. On my computer when I run pytest test_fastsrm.py everything works fine and I get no segmentation fault.

mihaic:
@hugorichard, it might be related to running under MPI. Could you please try `pr-check.sh` as well on your computer?

hugorichard:
It is related to MPI. It also does not run on my computer. py.test works but if I do mpiexec -n 2 py.test it does not seem to work anymore.

hugorichard:
I think I know it is because of joblib. It does not support nested parallelization loops. If I remove the concerned tests then it works. Should I do that ?

mihaic:
If joblib does not support running under MPI (a reference would be helpful), then we should test it separately, indeed. One small issue is making sure the coverage report is correct. I will look into it next week, unless you have experience with Coverage.py.

hugorichard:
./pr-checks now works locally. It still does not work online though.

mihaic:
We are seeing problems with the Conda builds in PR #440 as well, so it is not related to the changes. I am investigating.

hugorichard:
Ok thanks!Le 21 oct. 2019 6:29 PM, Mihai Capotă <notifications@github.com> a écrit :We are seeing problems with the Conda builds in PR #440 as well, so it is not related to the changes. I am investigating.

hejiaz:
@hugorichard The new changes look good. I will accept it after Mihai fixed the Conda build problem.

hugorichard:
This last commit just contains an updated example notebook that is consistent with the API (I also included wget as an example requirement to get a nice progress bar when people download the data used for the example but I can remove it if necessary).

mihaic:
@hugorichard, we fixed the build problem. Please update your PR branch from master.

mihaic:
@hejiaz, all tests are passing.

mihaic:
@hugorichard, thank you very much for your contribution! We look forward to continuing the collaboration.

hugorichard:
@mihaic You are very welcome. Thanks for your help and thanks to the reviewers !

PR #425, opened 2019-06-17
dev: Add brainiak-tutorials to Dockerfile

mihaic:
Move to Ubuntu 18.04 for Python 3.6, required for tutorials.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,11 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
+ARG DEBIAN_FRONTEND=noninteractive
 # Group 1 must be synced with README
 # Group 2 must be synced with requirements for examples
 # Group 3 must be synced with rest of Dockerfile
 # Group 4 is optional
+# Group 5 must be synced with requirements for tutorials
 RUN apt-get update && apt-get install -y \
     build-essential \
     libgomp1 \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -23,22 +25,27 @@ RUN apt-get update && apt-get install -y \
...
+    && python3 -m pip install --user -U . \
+    && python3 -m pip install --user -U -r tutorials/requirements.txt \
+    && for example in examples/*/requirements.txt; \
         do python3 -m pip install --user -U -r $example ; done
 
 RUN echo PATH=\"\$HOME/.local/bin:\$PATH\" >> $HOME/.profile \
     && echo "shell -bash" >> ~/.screenrc
 
-EXPOSE 8888
+EXPOSE 8899
```
gdoubleyew: Didn't we want to stay with port 8888 because that is what Jupiter prints as the URL when it starts? I remember users getting confused by that.

Dockerfile
```diff
@@ -23,22 +25,27 @@ RUN apt-get update && apt-get install -y \
...
 
 ENTRYPOINT ["/bin/bash", "-l"]
+CMD ["tutorials/run_jupyter_docker.sh"]
```
gdoubleyew: Do we want the tutorials to automatically start? Seems like most of the time the user of brainiak docker will not be using the tutorials.

Dockerfile
```diff
@@ -23,22 +25,27 @@ RUN apt-get update && apt-get install -y \
...
 
 RUN set -e \
-    && python3 -m pip install --user -U pip \
-    && python3 -m pip install -U brainiak \
-    && python3 -m pip download --no-deps --no-binary :all: brainiak \
-    && export BRAINIAK_VERSION=$(basename brainiak-* .tar.gz | cut -b 10-) \
-    && tar -xf brainiak-*.tar.gz \
-    && for example in brainiak-$BRAINIAK_VERSION/examples/*/requirements.txt; \
+    && python3 -m pip install --user -U "pip<10" \
+    && python3 -m pip install --user -U . \
```
gdoubleyew: For clarity I'd prefer to have WORKDIR /mnt and
&& python3 -m pip install --user -U brainiak \
&& python3 -m pip install --user -U -r brainiak/tutorials/requirements.txt \
&& for example in brainiak/examples/*/requirements.txt; \

Dockerfile
mihaic: This is consistent with the script that we run automatically, so Jupyter will print the right thing:
https://github.com/brainiak/brainiak-tutorials/blob/master/tutorials/run_jupyter_docker.sh#L2

Dockerfile
mihaic: I think the tutorials are a great way to start for all users of BrainIAK. At the same time, we should add instructions to the website about starting a shell (hint: add `-c bash` at the end of the `docker run` command).

Dockerfile
mihaic: From the user perspective, I think WORKDIR should be `brainiak`, so users can easily find the tutorials and examples. As for clarity in the Dockerfile for developers, I think developers are used to working inside the `brainiak` directory.

PR #432, opened 2019-08-08
DOC:Add example in docstring of load_images_from_dir

### LINKED ISSUE #369: io: Add list example for load_images_from_dir
Each time MaskedMultiSubjectData is called it requires the images of interest to be loaded and masked, otherwise a "StopIteration:" error appears, even if the images have been previously loaded. A helpful optimization would be to allow images to be loaded (load_images) and masked once initially and then to be able to call MaskedMultiSubjectData multiple times thereafter without having to reload/re-mask the images for each call of MaskedMultiSubjectData. A possibility is to update the load_images function so that the object created can be operated on more than once.

sameshl:
closes #369

### DIFF
--- a/brainiak/io.py
+++ b/brainiak/io.py
@@ -60,7 +60,7 @@ def load_images_from_dir(in_dir: Union[str, Path], suffix: str = "nii.gz",
 
     Examples
     --------
-    images = list(io.load_images_from_dir("face_scene", "bet.nii.gz"))
+    >>> images = list(io.load_images_from_dir("face_scene", "bet.nii.gz"))
     """
     if isinstance(in_dir, str):
         in_dir = Path(in_dir)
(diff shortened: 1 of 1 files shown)


buildbot-princeton:

buildbot-princeton:

mihaic:
Jenkins, OK to test.

brainiak/io.py
```diff
@@ -57,6 +57,10 @@ def load_images_from_dir(in_dir: Union[str, Path], suffix: str = "nii.gz",
     ------
     SpatialImage
         Image.
+
+    Examples
+    --------
+    images = list(io.load_images_from_dir("face_scene", "bet.nii.gz"))
```
mihaic: Let's use the leading `>>>` from the doctest format.

mihaic:
Thank you for your contribution, @sameshl.

sameshl:
Glad to help! @mihaic

PR #437, opened 2019-09-15
Drift update

CameronTEllis:
Update to how drift is calculated and used. The new default for drift drops off power on the basis functions systematically. Drift is also scaled separately from the brain noise since that scale is based on a detrended version of drift


CameronTEllis:
@mihaic I didn't add any code that I didn't subsequently test. Is there any way in which this can be fixed without me having to add tests to other code?

mihaic:
How about `if basis_funcs == 0`?
https://codecov.io/gh/brainiak/brainiak/pull/437/diff#D1-1617

mihaic:
@CameronTEllis, I fixed the latest failure in PR #439. Update your branch from master.

CameronTEllis:
@mihaic Did I do something wrong?

mihaic:
Jenkins, retest this please.

mihaic:
That was my fault. The cleanup script was not preparing the repository properly before the test, so there were leftover files from another PR causing the error. I changed it in Jenkins.

CameronTEllis:
@lcnature Do you have time to review this?

brainiak/utils/fmrisim.py
```diff
@@ -1595,6 +1603,57 @@ def _generate_noise_temporal_drift(trs,
         phase = (timepoints / (trs - 1) * cycles * 2 * np.pi) + phaseshift
         noise_drift = np.sin(phase)
 
+    elif basis == 'cos_power_drop':
+
+        # Specify when each
```
lcnature: Specify when each TR happens?

brainiak/utils/fmrisim.py
```diff
@@ -1595,6 +1603,57 @@ def _generate_noise_temporal_drift(trs,
...
+        timepoints = np.linspace(0, trs - 1, trs) * tr_duration
+
+        # Specify the other timing information
+        duration = trs * tr_duration
+
+        # How bases do you have
```
lcnature: How many bases do you have?

brainiak/utils/fmrisim.py
```diff
@@ -1595,6 +1603,57 @@ def _generate_noise_temporal_drift(trs,
...
+            timepoint_phase = (timepoints / duration * np.pi * basis_counter)
+
+            # In radians, what is the value for each time point
+            timepoints_basis = timepoint_phase + random_phase
+
+            # Store the drift from this basis func
+            noise_drift[:, basis_counter - 1] = np.cos(timepoints_basis)
+
+        # Function to return the drop rate for the power of basis functions
+        def power_drop(r, L, F):
```
lcnature: What does this each line do? It may help to explain why th 4 is there, and why the power can both a number without unit ( 1 - r ** (4 * L / F) ) and a number with unit (of time): ( denominator = 1 - r ** (4 * L) )

brainiak/utils/fmrisim.py
```diff
@@ -2744,6 +2800,22 @@ def generate_noise(dimensions,
...
+    if noise_dict['drift_sigma'] != 0:
+        # Calculate the drift time course
+        noise = _generate_noise_temporal_drift(len(stimfunction_tr),
+                                               tr_duration,
+                                               )
+        # Create a volume with the drift properties
+        volume = np.ones(dimensions[:3])
+
+        # Combine the volume and noise
+        drift_noise = np.multiply.outer(volume, noise) * noise_dict[
```
lcnature: For consideration of future improvement: I think this line assumes that all voxels have the same time series of drift with the same magnitude. So I think it can capture what is removed by global signal regression. But I think in reality the drift may be of different sizes across the brain.

brainiak/utils/fmrisim.py
```diff
@@ -1595,6 +1603,57 @@ def _generate_noise_temporal_drift(trs,
         phase = (timepoints / (trs - 1) * cycles * 2 * np.pi) + phaseshift
         noise_drift = np.sin(phase)
 
+    elif basis == 'cos_power_drop':
```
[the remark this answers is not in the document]
CameronTEllis: Thanks so much for looking at this code @lcnature, huge help. 

Regarding this, I just want to clarify what I think the code is doing and then if this doesn't jive with what you think should be done then let me know. `timepoints` is a vector of values in radians, describing the phase of the drift at each TR. In the logic of this function, for each basis function the frequency of the drift is reduced. For instance if the length of the run was 10x the period then the `timepoints` would range from 0 to 20*pi. For the first basis function I would simulate a cos function that has 10 cycles. Then if there were a second basis function I would simulate a cos function with 5 cycles, and so on. The randomness I am adding with `np.random.rand() * np.pi * 2` is simply to randomise the phase for each basis function. I am not sure what putting the np.random.rand() inside the bracket would achieve, but let me know if I am wrong about this

brainiak/utils/fmrisim.py
CameronTEllis: Thanks for the catc

brainiak/utils/fmrisim.py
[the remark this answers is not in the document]
lcnature: Hi @CameronTEllis Sorry I missed the fact that 2 pi was already multiplied to `timepoints` before. Then please ignore my comment on this line. What you do is correct.

brainiak/utils/fmrisim.py
CameronTEllis: Yeah this part was tricky, Chris did the heavy lifting with the math. Hopefully I have explained it better now.

brainiak/utils/fmrisim.py
CameronTEllis: Yeah completely agree, that is true. I thought about using a gaussian like with the other sources of noise but that also wouldn't be a good approximation. So yes would be a nice feature to add in the future

CameronTEllis:
Jenkins, retest this please.

CameronTEllis:
@mihaic Why is Travis crashing for ruby?

brainiak/utils/fmrisim.py
```diff
@@ -1595,6 +1603,66 @@ def _generate_noise_temporal_drift(trs,
...
+
+        # Make a vector counting each TR
+        timepoints = np.linspace(0, trs - 1, trs) * tr_duration
+
+        # Specify the other timing information
+        duration = trs * tr_duration
+
+        # How bases do you have?
+        basis_funcs = int(np.floor(2 * duration))
```
lcnature: This comment is related to that about `power_drop` to help me understand the logic. From the line above, I see `duration` is the total duration of the scan in units of a second, probably. So `2*duration` should also be a quantity of time (in units of a second). Then this `basis_funcs` is used below at my next comment.

brainiak/utils/fmrisim.py
```diff
@@ -1595,6 +1603,66 @@ def _generate_noise_temporal_drift(trs,
...
+
+        noise_drift = np.zeros((timepoints.shape[0], basis_funcs))
+        for basis_counter in list(range(1, basis_funcs + 1)):
+            # What steps do you want to take for this basis function
+            random_phase = np.random.rand() * np.pi * 2
+
+            timepoint_phase = (timepoints / duration * np.pi * basis_counter)
```
lcnature: Here, `basis_counter` goes from 1 to `basis_funs`, which is twice the total duration of the scan, in the unit of a second. `timepoints` is equally spaced time points between 0 and `duration`. When it is divided by `duration`, the result is equally spaced numbers between 0 and 1. This times pi would result in equally spaced points between 0 and pi. But multiply this phase with `basis_counter` which is a duration seems a bit strange, to be. It won't mean something as a radius anymore, but would be added with `random_phase` and be used as an input to cosine function. I think the unit here may be somehow wrong. Likely, the `basis_funcs` should be something like the ratio between the lowest frequency of the fft of the time series to the Nyquist frequency (twice of the sampling frequency of the scanner) instead, in line 1615?

brainiak/utils/fmrisim.py
```diff
@@ -1595,6 +1603,66 @@ def _generate_noise_temporal_drift(trs,
...
+        sol = optimize.minimize_scalar(power_drop,
+                                       bounds=(0, 1),
+                                       method='Bounded',
+                                       args=(duration, period))
+
+        # Pull out the solution
+        r = sol.x
+
+        # Weight the basis functions based on the power drop off
+        basis_weights = r ** np.arange(basis_funcs)
```
lcnature: According to this line, I feel the `basis_funcs` is something that ranges from 1 to the *index* of the Nyquiest frequency. Assuming it is true, then now I suppose what you are doing is to let the weight of each frequency component to decay exponentially. At each consecutive frequency, you decay by `r`.

brainiak/utils/fmrisim.py
lcnature: Hi @CameronTEllis Thanks for the explanation below!
Sorry I am still a bit confused so I will try to express my understanding with my new comments at other places of the code.

brainiak/utils/fmrisim.py
```diff
@@ -1595,6 +1603,66 @@ def _generate_noise_temporal_drift(trs,
...
+            # Function to return the drop rate for the power of basis functions
+            # In other words, how much should the weight of each basis function
+            # reduce in order to make the power you retain be above a 0.99
+            # r is the power reduction rate which should be between 0 and 1
+            # L is the duration of the run in seconds
+            # F is period of the cycle in seconds
+
+            percent_retained = 0.99  # What is the percentage of drift retained
+            numerator = 1 - r ** (4 * L / F)  # Power of this period
+            denominator = 1 - r ** (4 * L)  # Power of all periods
```
lcnature: According to my comment on Line 1658, I now guess `r` is the reduction rate of the weight of each frequency component from its previous one. So the power it would be raised to would be a unit-less number, not a duration. But `L` is the duration of the scan, which seems to be should not be a power. I am also not clear why 4 is multiplied. If the amplitude is decreased by `r`, then we can say the power is decreased by `r^2`. So I can understand if 2 is used, but not 4.

brainiak/utils/fmrisim.py
```diff
@@ -1595,6 +1603,66 @@ def _generate_noise_temporal_drift(trs,
...
+        def power_drop(r, L, F):
+            # Function to return the drop rate for the power of basis functions
+            # In other words, how much should the weight of each basis function
+            # reduce in order to make the power you retain be above a 0.99
+            # r is the power reduction rate which should be between 0 and 1
+            # L is the duration of the run in seconds
+            # F is period of the cycle in seconds
+
+            percent_retained = 0.99  # What is the percentage of drift retained
+            numerator = 1 - r ** (4 * L / F)  # Power of this period
```
lcnature: In this line, at least `L / F` is unitless, corresponding to the number of periods within the duration. So that makes sense to serve as the power that `r` is raised to (see my comment below).
This to me looks like something as the amount of reduction of power from the lowest frequency to the highest frequency. If `demoninator` is something like the amount of reduction in the power of starting frequency (from some baseline), then you are trying to say that from this starting frequency to the highest frequency, we have only dropped to 0.99. Is this correct? This does not seem to resemble what I would imagine as "power you retain" although I am also not so clear of the meaning of "retain" here. But I think maybe I misunderstood some part. Please feel free to correct me.

mihaic:
@CameronTEllis, fixed with PR #443.

brainiak/utils/fmrisim.py
lcnature: Thanks. Now I think I may have got it. You are trying to make a power spectrum that decreases exponentially from the lowest frequency (inverse of double the duration of the scan, corresponding to when `basis_counter` equals 1) to the highest frequency (the sampling frequency of fMRI,  when `basis_counter` equals `basis_funcs`).

lcnature: But then I am confused with the description "the power drop is meant to drop off to 99% by basis_funcs at duration, with the remaining power being captured in the subsequent basis_funcs duration->2*duration." Can I interpret it as, if I am integrating an exponential decaying curve from 0 up to certain number (duration) here, the integral is 99% of the integral from 0 to 2*duration? If this is the case, I would imagine the power of the numerator and denominator in the function `power_drop` should differ in a ratio of 2, instead of a ratio of `F`?

lcnature: Actually I think my second to last comment was incorrect. The highest frequency considered here is not the sampling frequency of fMRI, but a frequency which is set to the value of `duration`. This means that if our scan is 300 s long, we are talking about the highest frequency of 300 Hz (to see this, the total phase change of `timepoint_phase` from its first element to the last element is `2 * duration * pi` when `basis_counter` is `basis_funcs`. Theoretically there seems to be no restriction of what we set as the highest frequency to consider. But why do we make this choice? We never deal with 300 Hz signals in fMRI anyway. It seems to me the more natural choice is to set the highest frequency as the Nyquest frequency (2/TR). In this case, `basis_funcs` can be set as `duration / TR`.

brainiak/utils/fmrisim.py
lcnature: Hmm, let me make some guess to see if I am understanding correctly. In the denominator, `4 * L` is `2 * (2 * L)`, where `2 * L` is `2 * duration`. This would make sense given the codes above where you have set the highest `basis_counter`  to be `2 * duration`. And the  `4 * L / F` is essentially `2 * (1/F / (1/2L))`, where `1/F` is the frequency corresponding to the period you consider retaining 99% of the power below it, and `1/2L` is the lowest frequency in your basis set. So now both of these make sense to me. My only concern is whether setting the highest frequency as `2*L` makes sense. They are of different units to start with: `2L` is a quantity of time, choosing its number as a frequency seems strange to me.

brainiak/utils/fmrisim.py
lcnature: After having a second thought, I think now I understand what you mean by retain. It may be fine, but some more explanation may help anyone else reading the code. Maybe you can complete your original sentence as "how much should the weight of each basis function reduce in order to make the power you retain up to the frequency with its period as F be above 99% of the total power up to the highest frequency" (although I think we have different ideas of what should be set as the highest frequency here, see my other comments)

brainiak/utils/fmrisim.py
CameronTEllis: This is a really tricky point so I will do my best to work through it but I am not sure if I can give a satisfying answer. Firstly, I think you implied that `r` is affected by the choice of what `basis_funcs` is; this is not the case, only `duration` and `period` affect `r`. Hence how many `basis_funcs` we estimate will ultimately be non-influential to the `r`. Secondly, pragmatically, this choice doesn't matter because the drop off is such that the influence of any basis function after about the 10th is close to zero. For reference, for a typical `r`, the 10th function has a weight of 2e-2, 50th weight is 4e-9, 100th is 2e-17. So the difference between going to 300 or 600 is negligible. Thirdly, that said, it isn't so much about what practically matters but what is theoretically correct. The way I think we were thinking of it is that we wanted to compensate for Nyquist by making sure we had sufficient resolution at the frequency of the duration. Does this seem reasonable?

brainiak/utils/fmrisim.py
CameronTEllis: Glad it makes sense, will wait on the above comment to see if this still seems workable

brainiak/utils/fmrisim.py
CameronTEllis: Good suggestion, added it with some light rewording

brainiak/utils/fmrisim.py
lcnature: Yes. But if it were not the case that `basis_funcs` was set to `2 * duration`, I don't understand what the meaing of `4 * L` is as the power in the denominator. When I say `basis_funcs` infuence `r`, I mean that you take a frequency which equals `duration` as your upper frequency, and it is this frequency that influences `r`. I think I am mainly concerned that such a choice of the highest frequency does not make sense.

However, putting that aside, you are right that in most cases, when you use `4 * L` as the power (regardless of what meaning it is), your denominator would essentially be 1. And in fact, if you consider 99% as the target of power you retain up to the frequency corresponding to `period` out of the total power up to the frequency of *infinity*, then you can keep the definition of the `numerator` but let the `denominator` be 1. This way, you don't really need to do any fitting. 
You can directly solve the equation of `numerator = 0.99` to get `r`. There is also no more confusion of what needs to be in the `denominator`, and in this case I also agree that `basis_funcs` does not influence `r`.

CameronTEllis: @lcnature I was convinced of your argument by this last message and so reached out to Chris and he agreed. A small tweak he suggested, that I think you also thought of, is to base the maximum frequency on the number of TRs, not the duration of the run. He gives the example that if you had a 100s TR for a 200s run you wouldn't want to to simulate 200 basis funcs but instead only 2. The code change I made should accord with that. Moreover, I made a change to the `power_drop` function in order to deal with the new `basis_funcs`. However, I kept the denominator as a variable rather than 1. Although practically for most cases you are right that it will be equal to 1, for short runs (e.g. <100s) this value will be non-trivially below 1. Plus, including that optimization is a quick step done only once per simulation.

lcnature: Thanks @CameronTEllis . It is better now. But I still think that if you don't change the denominator to 1, the power in the denominator would be something like `2 * trs` - the power should be unit-less (I know practically they both make the denominator so close to 1 with almost no difference). I was almost convinced that the 4 in the original code makes sense but I am uncertain now. @cbaldassano thoughts? Should it be `numerator = 1 - r ** (4 * L / F)` for `2 * (1/F / 1/2L)` and `denominator = 1 - r ** (2 * trs)` for `2 * 1/2tr / 1/2L`, which is `2 * trs`? Here, `1/F` is the frequency of "retaining", `1/2tr` is the Nyquist frequency and `1/2L` is the lowest non-zero frequency in the spectrum? I am a bit uncertain whether it should be `/1L` or `1/2L`.

cbaldassano: @lcnature  is right that the units aren't right there. The reasoning for having the sum be up to L isn't based on the Nyquist freq (though it is related), it is because that is the number of basis functions in the DCT. For example see the top of [pg 2 here](https://link.springer.com/content/pdf/10.1007%2F978-1-4471-6684-9_20.pdf) - u indexes the timepoints, m indexes the basis functions. The twos are coming from the geometric series equation, since power is proportional to r**2 (so these terms could be written, for example, (r**2)^(L/F)).

I believe the numerator is correct as is but the denominator needs to be put back into # of TRs units, e.g.

```
            numerator = 1 - r ** (2 * L / F)  # Power of this period
            denominator = 1 - r ** (2 * L / tr_duration)  # Power of all periods
```

CameronTEllis: Thanks for this clarification @cbaldassano , it helped my understanding. Because of it I added, these changes, some documentation, an example and a check to ensure that `tr_duration` < `period`. However, reading this did raise a concern about another line of code https://github.com/brainiak/brainiak/blob/e21342d26764f245765378e59edb5f3c90985d40/brainiak/utils/fmrisim.py#L1679 (`basis_weights = r ** np.arange(basis_funcs)`). This line is meant to find weights for each of the basis functions before they are all added together; however, based on my new understanding shouldn't this be: `basis_weights = r ** (2 * np.arange(basis_funcs))` in order to make power proportional?

cbaldassano: No, the power spectrum is the square of the coefficients. So the coefficients themselves are 1, r, r^2, r^3... and the power spectrum of the frequencies is 1, r^2, r^4, r^6...

CameronTEllis: Aha, okay thanks for that clarification. The code should then be ready to go, what do you think @lcnature ?

lcnature: Looks great. I am glad to approve it!

cbaldassano:
Thanks @lcnature for your careful review of this!

CameronTEllis:
@mihaic Failed because of error in the testing of FCMA, which I didn't edit. I think this was a statistical fluke, should I just retest?

mihaic:
It's an issue we are working on, issue #448.

CameronTEllis:
@mihaic does this resolve issue #448 suitably? Made the edits on this branch for simplicity.

mihaic:
@CameronTEllis, it looks to me that the existing test was looking for some specific output, not for random output. Did I misunderstand that?

CameronTEllis:
@mihaic correct. The test generates random data and then runs an SVM with pseudo-random labels. The test Yida wrote is based on running this classifier once, checking what results it gets and then using those results to check that future tests get the same result. In this code there is some randomness in terms of how the CV is set, but this was controlled previously by setting the seed. However, I think the update added an additional call to the rng which means that the same code will not produce the same results. 

Regardless, the spirit of the test is whether the code outputs something of a certain shape and then asks whether it is at chance. I can't think of why getting the exact same result as before is necessary for testing the code: you could change the code in problematic ways and still (sometimes) get the same results or you could change the code in non-problematic ways (e.g. extra call to rng) and not get the same results. Moreover, this is not a standard we apply consistently in our tests because speaking from experience the tests in fmrisim do not require this.

mihaic:
@CameronTEllis, the error I see is about the length of the results, not the contents. I would rather pin scikit-learn so we can resume merging and later try harder at preserving the existing comparison. What do you think?

CameronTEllis:
@mihaic The length discrepancy is a misdirection. The code runs a searchlight on a mask and computes the classification accuracy in each voxel within the mask mask. If the voxel is outside of the searchlight it doesn't return a `None` it returns a `0`. Also, if the classification gets 0/8 correct then it will also produce a `0`. Hence, in the output there are 2 reasons for having `0`s, one of which depends on the classification performance. The previous test ignores all `0`s and only considers the classification accuracy for the non-zero values; so if there were a different number of 0/8 classifications in the whole brain volume then there will be a different number of outputted results of non-zero values. As mentioned, the classifier was affected by this update so that changed the number of 0/8 classifications.

tests/fcma/test_mvpa_voxel_selection.py
```diff
@@ -43,6 +43,8 @@ def test_mvpa_voxel_selection():
                 output.append(int(8*tuple[1]))
         expected_output = [6, 6, 5, 5, 5, 5, 5, 5, 5, 4, 4, 4, 4, 4,
                            4, 4, 4, 3, 3, 3, 3, 3, 2, 2, 2, 1]
+
+        # Check that the mean accuracy with noise is close to chance
```
mihaic: @CameronTEllis, you missed a line when you undid the test change. Next time, try `git revert`.

PR #440, opened 2019-10-18
Allowing the `sl_rad` parameter in searchlight to be 0

lcnature:
There are use cases in which we want to do the same processing on all voxels and want to use MPI to parallelize this. In this case, the searchlight has the size of 1, and it corresponds to `sl_rad = 0` when initializing `SearchLight`. But the current implementation will distribute an empty array to each worker if we set `sl_rad = 0`. The consequence is that nothing gets computed for each voxel. This PR fixes this bug.

### DIFF
--- a/brainiak/searchlight/searchlight.py
+++ b/brainiak/searchlight/searchlight.py
@@ -155,6 +155,8 @@ class Searchlight:
     """
     def __init__(self, sl_rad=1, max_blk_edge=10, shape=Cube,
                  min_active_voxels_proportion=0):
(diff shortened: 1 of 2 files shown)


mihaic:
I hope the Conda build problem is intermittent and it will be gone on Monday. :) Will review then.

CameronTEllis:
Happy to review this once those issues are resolved

lcnature:
Awesome! @CameronTEllis you are welcome to review it as well. The changes are minor.

mihaic:
@CameronTEllis, you can review; the errors are unrelated.

Issue #107, opened 2016-09-20
Add guidelines for duration of tests ran on PRs


mihaic:

mihaic:
Travis allows 50 minutes:
https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts

mihaic:
Unless someone objects, I will add a 1-minute maximum test duration per package on a laptop. Considering that Travis allows 50 minutes, this should give us some room for expansion.

PR #441, opened 2019-10-29
dev: Increase Travis output wait

mihaic:


mihaic:
Documentation for `travis_wait`:
https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

mihaic:
I tested a couple of values before submitting this PR. The default without this PR is 10 minutes; 30 minutes is too short as you can see in this failed build of my fork:
https://travis-ci.org/mihaic/brainiak/builds/604666246

Travis has an overall timer of 50 minutes as well and we must allow time for testing. See also the fixed issue where we discussed test durations:
https://github.com/brainiak/brainiak/issues/107

I believe a recent change in `conda-build` increased the build time, so I opened an issue to investigate that:
https://github.com/brainiak/brainiak/issues/442

Issue #446, opened 2019-12-03
FastSRM fails when the number of sessions is > 2


hugorichard:
There is a bug in my implementation causing FastSRM to fail when the number of sessions is > 2.
I'll submit a pull request for this.

mihaic:
Fixed in PR #447.

PR #447, opened 2019-12-03
Fix Fastsrm issue #446

hugorichard:
Solves #446 

I also added a perturbation when computing the basis so that DetSRM and FastSRM without an atlas matches exactly when low_ram mode is disabled.


hugorichard:
I don't understand the failure. It does not seem to be related to my PR:

=================================== FAILURES ===================================

__________________________ test_mvpa_voxel_selection ___________________________

    def test_mvpa_voxel_selection():

        data = prng.rand(5, 5, 5, 8).astype(np.float32)

        # all MPI processes read the mask; the mask file is small

        mask = np.ones([5, 5, 5], dtype=np.bool)

        mask[0, 0, :] = False

        labels = [0, 1, 0, 1, 0, 1, 0, 1]

        # 2 subjects, 4 epochs per subject

        sl = Searchlight(sl_rad=1)

        mvs = MVPAVoxelSelector(data, mask, labels, 2, sl)

        # for cross validation, use SVM with precomputed kernel

    

        clf = svm.SVC(kernel='rbf', C=10)

        result_volume, results = mvs.run(clf)

        if MPI.COMM_WORLD.Get_rank() == 0:

            output = []

            for tuple in results:

                if tuple[1] > 0:

                    output.append(int(8*tuple[1]))

            expected_output = [6, 6, 5, 5, 5, 5, 5, 5, 5, 4, 4, 4, 4, 4,

                               4, 4, 4, 3, 3, 3, 3, 3, 2, 2, 2, 1]

>           assert np.allclose(output, expected_output, atol=1), \

                'voxel selection via SVM does not provide correct results'

tests/fcma/test_mvpa_voxel_selection.py:46:

hugorichard:
Actually this error also occurs in master...

mihaic:
Indeed, the failure is unrelated. See issue #448.

mihaic:
@hejiaz, could you please review? The testing should be fixed on master.

CameronTEllis:
Happy to take the lead on reviewing this, unless you have already started @hejiaz

brainiak/funcalign/fastsrm.py
```diff
@@ -926,7 +926,9 @@ def _compute_subject_basis(corr_mat):
     or shape=[n_components, n_supervoxels]
         basis of subject or reduced_basis of subject
     """
-    U, _, V = scipy.linalg.svd(corr_mat, full_matrices=False)
+    perturbation = np.zeros(corr_mat.shape)
```
CameronTEllis: It would be good to add a comment as to why you needed to perturb the correlation matrix before running the SVD

brainiak/funcalign/fastsrm.py
```diff
@@ -977,7 +979,7 @@ def fast_srm(reduced_data_list,
         for j in range(n_sessions):
             timeframes_slices.append(
                 slice(current_j, current_j + len(reduced_data_list[0, j])))
-            current_j = len(reduced_data_list[0][j])
+            current_j += len(reduced_data_list[0][j])
```
CameronTEllis: It would be helpful to add to the comments for this function to describe how `reduced_data_list` differs from what SRM needs and hence why you need to reshape it like you do

brainiak/funcalign/fastsrm.py
hugorichard: This is for consistency reasons with respect to previous implementation of SRM. Running FastSRM with no atlas is now equivalent to running DetSRM which is a good thing (I added a comment in the code)

brainiak/funcalign/fastsrm.py
hugorichard: As FastSRM API is more permissive than DetSRM or ProbSRM because the input data can maintain the division by session usually seen in fMRI datasets. This is why we need to concatenate the data across sessions when performing DetSRM. The code enables FastSRM to revert this operation in order to get a consistent API. (I added a comment in the code).

hugorichard:
There is probably a configuration issue because the tests do not pass anymore @mihaic

mihaic:
Yes, there are some issues because of dependencies. I will let you know when I fix them so you can merge the fixes from master.

mihaic:
@hugorichard, ready to merge as soon as you update the pull request branch. Sorry for the delay. We still have not fixed all the issues, but are now allowing merges with certain failures.

hugorichard:
@mihaic I just pulled the code from master.

Issue #448, opened 2019-12-04
fcma: test_mvpa_voxel_selection fails with scikit-learn 0.22


mihaic:
See https://travis-ci.org/brainiak/brainiak/jobs/620219411

Pinning to <0.22 works, but we should fix the underlying issue.

CameronTEllis:
Is anyone working on this? I can take a crack at it if no one else is

mihaic:
It's all yours!

mihaic:
@CameronTEllis, I think it is better to contine the conversation from PR #437 here. Your explanation makes sense; it is likely the length difference is caused by a change in accuracy. Nevertheless, all but one of the accuracy values are the same. Therefore, I think it should be possible to preserve the accuracy value comparison in the test.

CameronTEllis:
Indeed the difference between the results are similar but that is incidental as far as I can tell. Again, I am not sure why the value comparison is the appropriate test because if so it sets an extreme standard that should affect all of our tests. If it is necessary to include this test then we can put in an if statement that deals with this different computation. Other than that I am unsure how to proceed

mihaic:
You are right, we have no standard for tests requiring such comparisons. However, I see no harm keeping this existing test, as long as it is working, and it seems close to working to me.

My proposal, as written in the issue description, is to pin scikit-learn until we figure out what is happening so we can release BrainIAK v0.10.

We can even give ourselves a deadline, let's say version BrainIAK v0.11, and remove the test if we cannot make it work by then.

@yidawang, perhaps you have more insight.

CameronTEllis:
That sounds reasonable then

PR #450, opened 2019-12-10
dev: Pin scikit-learn <0.22 to fix fcma test

mihaic:
See issue #448.

Also, pin NumPy in `setup_requires` as well.


mihaic:
Thanks. I created a milestone for v0.11 and added the issue there so we don't forget. Now we only need to remember to check the milestone before releasing. :)
https://github.com/brainiak/brainiak/milestone/6

PR #451, opened 2020-01-14
Split-merge

cbaldassano:
This pull request implements an improvement to the fitting procedure for the HMM, allowing for re-allocating events during fitting with a split-merge operation. This behavior is disabled by default so as not to break backwards compatibility and because it adds additional computation time, but can be enabled using the new split_merge flag in the constructor.

A draft of a blog post explaining the logic behind this change and some example fits can be found here: http://chrisbaldassano.com/Split-Merge.mhtml


cbaldassano:
Jenkins is failing with "salloc: error: Job submit/allocate failed: Unable to contact slurm controller (connect failure)" - looking through my emails I think slurm is down for maintenance at Princeton today, can a maintainer please restart the checks once it is back up?

mihaic:
@cbaldassano, thanks for the heads up. We are working of fixing the problems.

@qihongl, you can see on Travis that the unit tests are passing, so please disregard the failures and review as soon as possible.

tsy6:
I would be happy to review this! @cbaldassano @qihongl

brainiak/eventseg/event.py
```diff
@@ -489,3 +538,143 @@ def model_prior(self, t):
...
+        merge_corr = np.zeros(self.n_events)
+        merge_pat = np.empty((n_train, n_dim, self.n_events))
+        split_corr = np.zeros(self.n_events)
+        split_pat = np.empty((n_train, n_dim, 2 * self.n_events))
+        for i, sp in enumerate(seg_prob):  # Iterate over datasets
+            merge = np.zeros((sp.shape[0], sp.shape[1]))
+            split = np.zeros((sp.shape[0], 2 * sp.shape[1]))
+            cs = np.cumsum(sp, axis=0)
+            for e in range(sp.shape[1]):
+                mid = np.where(cs[:, e] >= 0.5)[0][0]
```
tsy6: This section was the most difficult one for me to follow, partly because of the use of variables named merge and split within the for loop. Merge_corr and merge_pat, for instance, are quite easy to grasp so if there is something you could add to make the purpose of this variable more obvious, I think it would help.

brainiak/eventseg/event.py
```diff
@@ -489,3 +538,143 @@ def model_prior(self, t):
...
+                merge[:, e] = sp[:, e:(e + 2)].mean(1)
+            merge_pat[i, :, :] = X[i].dot(merge)
+            split_pat[i, :, :] = X[i].dot(split)
+        merge_pat = np.mean(merge_pat, axis=0)
+        split_pat = np.mean(split_pat, axis=0)
+        for e in range(self.n_events):
+            split_corr[e] = np.corrcoef(mean_pat[:, e],
+                                        split_pat[:, (2 * e):(2 * e + 2)],
+                                        rowvar=False)[0, 1:3].max()
+            merge_corr[e] = np.corrcoef(merge_pat[:, e],
```
tsy6: I'm curious -- why take the maximum split correlation and the minimum merge correlation? Why not average?

brainiak/eventseg/event.py
```diff
@@ -489,3 +538,143 @@ def model_prior(self, t):
...
+                                        rowvar=False)[0, 1:3].min()
+        merge_corr = merge_corr[:-1]
+
+        # Find best merge/split candidates
+        best_merge = np.flipud(np.argsort(merge_corr))
+        best_merge = best_merge[:self.split_merge_proposals]
+        best_split = np.argsort(split_corr)
+        best_split = best_split[:self.split_merge_proposals]
+
+        # For every pair of merge/split candidates, attempt the merge/split
```
tsy6: All of this looks good to me. Very neat intuition and way of implementing a solution to unequal-sized events!

brainiak/eventseg/event.py
cbaldassano: I moved and added comments to put them closer to the specific lines of code the referenced and add some more details, and changed the merge/split variables to be more clear.

brainiak/eventseg/event.py
cbaldassano: The goal is to select the merge and split combination that is most likely to increase the log likelihood, so we want to find the pair of events that is most similar to merge, and the split halves of an event that are most dissimilar to split - I've added more comments now.

brainiak/eventseg/event.py
cbaldassano: Thanks! Let me know if there is anything that is still not clear after my commit.

cbaldassano:
Thanks @tsy6! @qihongl if you agree that this can be merged then I think it is ready to go - not sure what is causing the linux-specific build error, the error log seems to be related to some kind of compiler misconfiguration (unrelated to this code)

mihaic:
Indeed, the build error was an unrelated. Should be fixed.

@cbaldassano, could you please also write a blurb for the release notes? See this old one if you need to refresh your memory on how they work:
https://github.com/brainiak/brainiak/blob/fe604afbf1b919fb9fe4ce9c5e2c8bede94ec3ef/docs/newsfragments/353.feature

cbaldassano:
Done!

mihaic:
Thank you for your contributions, @cbaldassano and @tsy6!

PR #457, opened 2020-03-12
Add rt fmrisim example

CameronTEllis:
Add a function for generating simulated data that might be compatible with the Real Time code.

Also provide a warning if inappropriate data is inappropriately passed to convolve_hrf


examples/utils/fmrisim_real-time_generator.py
```diff
@@ -0,0 +1,402 @@
...
+import random
+import argparse
+import datetime
+import nibabel  # type: ignore
+import numpy as np  # type: ignore
+import pydicom as dicom
+from brainiak.utils import fmrisim as sim  # type: ignore
+import sys
+
+sys.path.append(os.getcwd())
```
mihaic: Do we need this line?

CameronTEllis: That is a good point, not sure why I added that, removing it now

PR #458, opened 2020-03-18
Add Dicom header fields to allow dcm2nii

gdoubleyew:
I added some Dicom header fields to get dcm2nii working, these include having a different timestamp and instance number for each image. There are still warnings when dcm2nii runs that I couldn't get rid of, namely it complains that the metadata group 2 byte length is not specified. After trying to fix that warning for awhile I've concluded that is something pydicom.save_as() would need to handle because it would know how many bytes the metadata section is at write time.

Commands used:
python fmrisim_real-time_generator.py -i ./sim_parameters -o /tmp/gen -n 100 --saveAsDicom
dcm2niix -a y /tmp/gen/

### DIFF
--- a/examples/utils/fmrisim_real-time_generator.py
+++ b/examples/utils/fmrisim_real-time_generator.py
@@ -38,6 +38,8 @@ import pydicom as dicom
 from brainiak.utils import fmrisim as sim  # type: ignore
 import sys
 
+script_datetime = datetime.datetime.now()
+
 def generate_ROIs(ROI_file,
                   stimfunc,
                   noise,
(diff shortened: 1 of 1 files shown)


gdoubleyew:
Oh, I should also mention, there was a bug regarding directory names. The fmrisim_dir and data_dir weren't getting set from the args parser fields. Sometimes they were called inputDIr and outputDir so I consolidated on using those names, and added os.path.join in case the directory separator '/' wasn't provided in the user argument.

CameronTEllis:
This looks great, no suggestions. I won't approve it though until we decide where to put this script.

PR #464, opened 2020-05-20
Added more values to the defaults

CameronTEllis:
Defaults added both within the argparser as well as outside of it. This will make the input of the data_dict into generate_data optional

### DIFF
--- a/brainiak/utils/fmrisim_real_time_generator.py
+++ b/brainiak/utils/fmrisim_real_time_generator.py
@@ -244,7 +244,8 @@ def _write_dicom(output_name,
 
 def _get_input_names(data_dict):
     """Get names from dict
-        Read in the data_dict to return the relevant file names
+        Read in the data_dict to return the relevant file names. Will also
+        add the default values if trDuration, isi, or burn_in haven't been set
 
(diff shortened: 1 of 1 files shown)


CameronTEllis:
@mihaic this is crashing because of other code I didn't change. Did something change in pytest to make it not like `l` as a variable name?

mihaic:
It's probably the new release of Flake8. Looking into it.

mihaic:
@CameronTEllis, the errors should go away if you get the latest commit from master.

CameronTEllis:
@gdoubleyew Can you review this please to see if it is suitable?

mihaic:
@gdoubleyew, in case you are wondering about the coverage failure, I think it is a fluke, so please ignore it.

brainiak/utils/fmrisim_real_time_generator.py
```diff
@@ -244,7 +244,8 @@ def _write_dicom(output_name,
 
 def _get_input_names(data_dict):
```
gdoubleyew, line 245: maybe we could rename this function to _get_input_defaults() ?

PR #469, opened 2020-07-13
Add a few missing defaults to get_input_defaults() in fmrisim_real_time_generator.py 

gdoubleyew:


brainiak/utils/fmrisim_real_time_generator.py
```diff
@@ -287,41 +288,55 @@ def _get_input_defaults(data_dict):
...
+    if 'scale_percentage' not in data_dict:
+        data_dict['scale_percentage'] = 0.5
+    if 'multivariate_pattern' not in data_dict:
+        data_dict['multivariate_pattern'] = False
+    if 'different_ROIs' not in data_dict:
+        data_dict['different_ROIs'] = False
+    if 'save_dicom' not in data_dict:
+        data_dict['save_dicom'] = False
+    if 'save_realtime' not in data_dict:
+        data_dict['save_realtime'] = False
```
mihaic, line 339: I was thinking that we could build the dictionary the other way around: start with a default `dict` and `update` it with the user data in `data_dict`. It would be easier to maintain and no trigger the C901 complexity error.

gdoubleyew: I see. One thing to consider is that the generate_data() function can be called from main(), in which case it will add the args params to the data_dict, but it can also be called directly from another python script (how I use it in the rt-cloud sample) in which case the passed in data_dict will just contain the settings they want to modify. So I guess you are suggesting that generate_data() start with a default_dict and then merge whatever dict is passed into it with the default_dict, is that right?

mihaic: That's right.

CameronTEllis:
Hi Grant, good work on this. Things look good but I noticed a couple issues. Do you want me to give them to you to bundle with this request or to follow it up when you have a commit that gets past Travis?

gdoubleyew:
How about give it to me now and if it seems too much we can follow it up with another commit.

CameronTEllis:
Okay great, so the changes in addition to what you made are just:

Remove references to inputDir on lines 5, 9, and 347-348 
Change line 385 to be: `data_dict['noise_dict'] = noise_dict`

mihaic:
@CameronTEllis, if you are waiting for the tests to review, you should know the tests passed. Not sure why GitHub does not update the status.

PR #474, opened 2020-07-28
Update the doc string to give more details for the noise_dict

### LINKED ISSUE #470: Description of how drift_sigma is used is wrong in fmrisim
In the doc strings it says that all noise types with the suffix `_sigma` are mixed together to contribute to the total of the temporal variance of noise. However, this is not true for `drift_sigma` in the latest version of fmrisim. Docstrings need to be updated to accommodate this nuance.

CameronTEllis:
Also make a test to ensure that the signal_function and noise_function are the same size in compute_signal_change.

Fixes #470


mihaic:
@CameronTEllis & @gdoubleyew, remember that you can also merge PRs.

Issue #449, opened 2019-12-08
isc: two-sample permutation test fails for pairwise ISCs


snastase:
When computing pairwise ISCs, the output is a condensed triangle vector (via `squareform`) with length `N * (N - 1) / 2`, where `N` is the number of subjects. In the context of two separate  (non-paired) groups, ISCs are computed separately and the resulting condensed vectors are v-stacked (note that there may not be the same number of subjects in each group). When passing this stacked pairwise ISC array for two groups to `permutation_isc`, the `_check_isc_input` doesn't handle the stacked groups properly. It assumes there's one group and breaks because that stacked array likely won't have the correct number of elements for `squareform`. I think most of the rest of the permutation test handles this situation properly, but the logic of `_check_isc_input` needs to be adjusted, and we need a test confirms this is handled properly. Thanks to [@mcdoar9](https://github.com/mcdoar9) for catching this!

snastase:
We've confirmed that this is not really a bug; more that the documentation is unclear as to how two-sample pairwise ISCs are to be supplied to `permutation_isc`. PR #476 should clarify this, so I'm closing this.

PR #476, opened 2020-07-31
Update ISC permutation doc to clarify pairwise groups

snastase:
This addresses the confusion with issue #449 (via @mcdoar9). The documentation was unclear on how to supply pairwise ISCs for two groups to the permutation test. The documentation is now updated to make it more clear that pairwise ISCs must be computed on both groups at once (such that between-group ISCs are included in the pairwise ISC matrix) prior to this ISC matrix being supplied to `permutation_isc`.

### DIFF
--- a/brainiak/isc.py
+++ b/brainiak/isc.py
@@ -1060,8 +1060,8 @@ def permutation_isc(iscs, group_assignment=None, pairwise=False,  # noqa: C901
     leave-one-out approach, ISC values for two groups should be stacked
     along first dimension (vertically) and a group_assignment list (or 1d
     array) of same length as the number of subjects should be provided to
-    indicate groups. In the pairwise approach, pairwise ISCs should be
-    computed the across both groups at once; i.e. the pairwise ISC matrix
+    indicate groups. In the pairwise approach, pairwise ISCs should have
+    been computed across both groups at once; i.e. the pairwise ISC matrix
     should be shaped N x N where N is the total number of subjects across
     both groups, and should contain between-group ISC pairs. Pairwise ISC
     input should correspond to the vectorized upper triangle of the square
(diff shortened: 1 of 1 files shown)


snastase:
Oops accidentally created this branch off the [`isc-boot-fix`](https://github.com/snastase/brainiak/tree/isc-boot-fix) branch at PR #475. Hopefully won't cause problems...?

CameronTEllis:
Great, Tristan volunteered for this

snastase:
@tristansyates good point! I fixed the tense. I think I'll leave the somewhat redundant input shape descriptions... the earlier one is intended to clarify that pairwise ISCs should have been computed in such a way as to produce a distance matrix containing both groups; the latter more specifically describes the input shapes (which for the pairwise approach should be distance vectors, not square).

mihaic:
@snastase, the current changes look off. If you try to fix the code, start with a pull from GitHub because I used the update branch from master button.

snastase:
> @snastase, the current changes look off. If you try to fix the code, start with a pull from GitHub because I used the update branch from master button.

@mihaic—not sure I totally understand. Does something in particular look off to you? It looks correct to me... The only weirdness is that the changes from #475 are also part of the diff here. If we merge #475, this will only have a small diff—i.e. the only real diff in this PR is the two added lines 1070-1071 in the docstring.

mihaic:
I thought the main change of this PR was introduced in commit
https://github.com/brainiak/brainiak/pull/476/commits/8e557d48c54d6118e678415bc0f60930f95fa1b7

I thought that change was missing in the current diff. However, I noticed that change was actually merged with PR #477.

If you only need lines 1070-1071, we can merge them right away if you remove the changes from #475. Or we can wait for #475.

Issue #466, opened 2020-06-05
Leave-one-out ISC permutations/bootstraps not valid


cbaldassano:
When using permutation_isc and bootstrap_isc with pairwise=False, the permutations are carried out by applying signflips, group exchanges, or bootstrap resampling to the correlation values of individual subjects with the rest of the group. For example, in the one-sample permutation each subject's LOO value is randomly flipped or not flipped:

https://github.com/brainiak/brainiak/blob/69a0c93cceea79215622221e7952f4b2262892aa/brainiak/isc.py#L938

I don't think this is correct. When using LOO ISC, each subject's data is used to compute ISC value (as part of the "group"), and flipping the sign of a subject's data would impact every LOO value for all subjects, not just the LOO value of that specific subject. I believe this problem is true for both the 1-sample and 2-sample permutation and bootstrap tests as they are currently coded.

The Chen2016 reference, as far as I can tell, only describes permutations and bootstraps in the pairwise case, and does not describe a valid way of doing this for the leave-one-out case. I can't think of a way to do this that would be valid, so my proposed solution would be to have these functions thrown an error if they are attempted to be called with LOO ISCs.

Sam, let me know if I am misunderstanding how these functions are working.

snastase:
@cbaldassano thanks for the comment! I'm trying to understand your question here... In the one-sample permutation test, the sign-flipping is applied to the correlation values resulting from the ISC computation (for both the pairwise and leave-one-out approaches). If you have `N = 30` subjects, at each permutation you'll get a 30-long vector of `-1`s and `+1`s (and for low `N`, we compute the `2**N` exact set of all permutations)—for both approaches. In the pairwise approach, the ISC computation results in an `N x N` correlation matrix (`N*(N - 1)/2` unique correlation values) and the sign-flipping is applied to the rows/columns of the matrix to achieve a subject-wise (not pair-or element-wise) permutation. In the leave-one-out approach, the ISC computation results in `N` correlation values and the sign-flipping is applied to these `N` values for subject-wise permutation. You're correct that in the leave-one-out approach, each of these `N` correlation values corresponds to a left-out subject's correlation with the average of the `N - 1` other subjects. But there's no flipping applied to the actual time series data feeding into the ISC computation (or components of the ISC computation like averaging the `N - 1` subjects)... I'm not sure I see *in principle* why this would cause a problem (but it might!). Do you find that behaves strangely in practice?

[...]

PR #477, opened 2020-07-31
Add 'side' option to randomization tests

snastase:
This adds a `side` (possible values: `two-tailed`, `right`, `left`) option to be passed to `p_from_null` in the ISC randomization tests for determining whether the p-value output is one- or two-tailed (based on conversations with @me-sh). Previously, `two-tailed` was hardcoded into the randomization tests. I've also set the default to `right`, since effectively all ISC analyses are seeking positive correlation (and negative correlations are pretty nonsensical in ISC analysis).

### DIFF
--- a/brainiak/isc.py
+++ b/brainiak/isc.py
@@ -692,7 +692,7 @@ def bootstrap_isc(iscs, pairwise=False, summary_statistic='median',
     ci_percentile : int, default: 95
          Percentile for computing confidence intervals
 
-    side : str, default: 'two-sided'
+    side : str
         Perform one-sided ('left' or 'right') or 'two-sided' test
 
     random_state = int or None, default: None
@@ -1109,7 +1109,7 @@ def permutation_isc(iscs, group_assignment=None, pairwise=False,  # noqa: C901
     n_permutations : int, default: 1000
         Number of permutation iteration (randomizing group assignment)
 
-    side : str, default: 'two-sided'
+    side : str
         Perform one-sided ('left' or 'right') or 'two-sided' test
 
     random_state = int, None, or np.random.RandomState, default: None
@@ -1305,7 +1305,7 @@ def timeshift_isc(data, pairwise=False, summary_statistic='median',
     n_shifts : int, default: 1000
         Number of randomly shifted samples
 
-    side : str, default: 'two-sided'
+    side : str
         Perform one-sided ('left' or 'right') or 'two-sided' test
 
     tolerate_nans : bool or float, default: True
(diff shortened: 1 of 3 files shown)


CameronTEllis:
Great looks good Sam, do you have any recommendations for reviewers?

snastase:
This is a fairly straightforward change, so probably anybody could review. Not sure @me-sh is available; maybe if @tristansyates is actively running ISCs right now, they could eyeball this as well?

mihaic:
@snastase, does this fix #466?

snastase:
> @snastase, does this fix #466?

No, we're still working on #466. That one's a bit more tricky. I don't think the code is "wrong" per se; it's more of a statistical question about which sort of randomization yields a null distribution that best controls false positive rates.

mihaic:
Got it. We included #466 in the next release milestone. Do you have an estimate for agreeing on a way to close it?

brainiak/isc.py
```diff
@@ -1097,6 +1109,9 @@ def permutation_isc(iscs, group_assignment=None, pairwise=False,  # noqa: C901
     n_permutations : int, default: 1000
         Number of permutation iteration (randomizing group assignment)
 
+    side : str, default: 'two-sided'
+        Perform one-sided ('left' or 'right') or 'two-sided' test
```
mihaic, line 1113: The `side` default described in this docstring is not the default assigned in the function declaration.

More generally, we should not describe defaults in the docstrings because their presence in declarations makes them visible both in the Python built-in `help` function and the HTML docs:
```
python3 -c "import brainiak.isc; help(brainiak.isc.permutation_isc)"
```
https://brainiak.org/docs/brainiak.html#brainiak.isc.permutation_isc

brainiak/isc.py
```diff
@@ -1287,6 +1305,9 @@ def timeshift_isc(data, pairwise=False, summary_statistic='median',
     n_shifts : int, default: 1000
         Number of randomly shifted samples
 
+    side : str, default: 'two-sided'
+        Perform one-sided ('left' or 'right') or 'two-sided' test
+
```
mihaic, line 1310: Same comment about default.

brainiak/isc.py
```diff
@@ -1435,6 +1459,9 @@ def phaseshift_isc(data, pairwise=False, summary_statistic='median',
     n_shifts : int, default: 1000
         Number of randomly shifted samples
 
+    side : str, default: 'two-sided'
```
mihaic, line 1462: Same comment about default.

snastase:
Thanks @mihaic—good catch! I didn't realize this about defaults in the docstring... so should I remove the `default: x` statements from the docstrings throughout in this PR?

mihaic:
I think we should have dedicated PRs for such cleanups. For now, lets just not add more default descriptions, so just remove the ones part of the diff.

snastase:
Okay, I removed the defaults from the docstring for each occurrence of this `side` argument.

Issue #69, opened 2016-07-18
Expose logp for Bayesian models in brainiak


mshvartsman:
SRM and (H)TFA, and Bayesian RSA are all Bayesian models. It would be great if they exposed a method that produces a log-likelihood given some data and parameters. This would make easier a number of things: 

- Bayesian model comparison. 
- Decomposing the model structure from the estimation machinery for better reuse of either/both.

TuKo:
@mshvartsman , how would a log-likelihood function would help with model comparison here?
Can you give an expected usage example?

mihaic:
@mshvartsman, can you please reply to Javier?

mshvartsman:
For Bayesian model comparison I guess we would need marginal likelihood, not likelihood given data and parameters, so maybe this is not as useful on this front. 

For the latter situation I'm imagining something like the following: if I estimate a joint hyperprior over some latent cognitive features of behavioral data and latent factors I could alternate doing MAP estimation on the brain factor model, cognitive features, and the hyper-prior. I would be reusing the factor model likelihood but the rest of the structure would be different. 

I'm hoping to put together a simpler version (using PCA instead of a fancy factor model) as a proof of concept, so if you prefer we can close this and wait on it until I have a concrete thing I want to plug one of these factor models into to reopen.

lcnature:
This is a good idea as I also have plan to offer a model comparison capacity in Bayesian RSA. The simple way is to give BIC or AIC score. Doing cross-validation can be better. But fit() seems to be reserved for giving back labels for testing data instead of giving log-likelihood. What function name would be good?
But anyway I prefer to keep this as lower priority for my code as I have a few other things to add first.

PR #478, opened 2020-08-13
Add a score function for SRM

Kajiyu:
This adds a function on SRM which calculates the log-likelihood of test-subjects' data for model selection.
We followed a style of scikit-learn: https://scikit-learn.org/stable/modules/generated/sklearn.mixture.GaussianMixture.html#sklearn.mixture.GaussianMixture.score

Co-authored-by: @lcnature .

### DIFF
--- a/brainiak/funcalign/srm.py
+++ b/brainiak/funcalign/srm.py
@@ -579,6 +579,8 @@ class SRM(BaseEstimator, TransformerMixin):
         # Transform from new subjects
         _, voxels = _init_w_transforms(data, self.features, random_states,
                                        self.comm)
+        
+        # Compute the transform matricies for test-subjects
         w = []
         for subject in range(subjects):
             _w = self._update_transform_subject(data[subject], self.s_)
--- a/brainiak/funcalign/srm.py
+++ b/brainiak/funcalign/srm.py
@@ -585,7 +585,6 @@ class SRM(BaseEstimator, TransformerMixin):
             w.append(_w)
         
         x, mu, rho2, trace_xtx = self._init_structures(data, subjects)
-        shared_response = self.s_
         sigma_s = self.sigma_s_
 
         # Sum the inverted the rho2 elements for computing W^T * Psi^-1 * W
@@ -622,15 +621,14 @@ class SRM(BaseEstimator, TransformerMixin):
                                                 op=MPI.SUM)
         log_det_psi = np.sum(np.log(rho2) * voxels)
 
-        # Compute its trace
-        trace_sigma_s = samples * np.trace(sigma_s)
+        # Compute the log-likelihood
         ll_score = self._likelihood(
                 chol_sigma_s_rhos, log_det_psi, chol_sigma_s,
                 trace_xt_invsigma2_x, inv_sigma_s_rhos, wt_invpsi_x,
                 samples)
         
         # Add the constant term
-        ll_score -= 0.5*samples*voxels*subjects*np.log(2*np.pi)
+        ll_score -= 0.5*samples*np.sum(voxels)*np.log(2*np.pi)
         return ll_score
 
 
--- a/brainiak/funcalign/srm.py
+++ b/brainiak/funcalign/srm.py
@@ -551,6 +551,87 @@ class SRM(BaseEstimator, TransformerMixin):
 
         sigma_s = self.comm.bcast(sigma_s)
         return sigma_s, w, mu, rho2, shared_response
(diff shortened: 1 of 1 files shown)


lcnature:
This PR partially deals with #69 by adding logp for SRM.

mihaic:
@qihongl, could you please review?

qihongl:
Oops sorry about the delayed reply! @lcnature @Kajiyu  I didn't notice it! I will take a look in 2 weeks. 

btw is there a test added to this module?

lcnature:
Thanks @qihongl ! It looks like we did not add. I will maybe work on it this weekend or next week.

qihongl:
@lcnature Ah do you mean there are some new updates that haven't been added to this PR?

mihaic:
@lcnature, please remember about this pull request. It seems close to completion.

brainiak/funcalign/srm.py
```diff
@@ -552,6 +552,87 @@ def _srm(self, data):
...
+
+        data : list of 2D arrays, element i has shape=[voxels_i, samples]
+            Each element in the list contains the fMRI data of one subject.
+
+
+        Returns
+        -------
+
+        ll_score : Log-likelihood of test-subject's data
+        """
```
qihongl, line 569: Might be better to have an input validation here to make sure all `d` in `data` have the same number of samples? 

I noticed that later on line 616, you are skipping any `d` in `data` that is `None`. I was wondering if this can be done earlier, such a removing `None` from the very beginning, or warn the user that there is None in the `data` list? Feel free to decide what's more user-friendly.

brainiak/funcalign/srm.py
```diff
@@ -552,6 +552,87 @@ def _srm(self, data):
...
+
+        # Transform from new subjects
+        _, voxels = _init_w_transforms(data, self.features, random_states,
+                                       self.comm)
+
+        # Compute the transform matricies for test-subjects
+        w = []
+        for subject in range(subjects):
+            _w = self._update_transform_subject(data[subject], self.s_)
+            w.append(_w)
```
qihongl, line 587: not necessary but I think the following is slightly faster
```
w = [self._update_transform_subject(data[subject], self.s_) for subject in range(subjects)]
```

brainiak/funcalign/srm.py
```diff
@@ -552,6 +552,87 @@ def _srm(self, data):
...
+
+        x, mu, rho2, trace_xtx = self._init_structures(data, subjects)
+        sigma_s = self.sigma_s_
```
qihongl, line 590: not necessary but how about just use `self.sigma_s_` on line 597?

brainiak/funcalign/srm.py
```diff
@@ -552,6 +552,87 @@ def _srm(self, data):
...
+
+        # Sum the inverted the rho2 elements for computing W^T * Psi^-1 * W
+        rho0 = (1 / rho2).sum()
+
+        # Invert Sigma_s using Cholesky factorization
+        (chol_sigma_s, lower_sigma_s) = scipy.linalg.cho_factor(
+            sigma_s, check_finite=False)
```
qihongl, line 597: referred in a comment above

brainiak/funcalign/srm.py
```diff
@@ -552,6 +552,87 @@ def _srm(self, data):
...
+        inv_sigma_s_rhos = scipy.linalg.cho_solve(
+            (chol_sigma_s_rhos, lower_sigma_s_rhos),
+            np.identity(self.features), check_finite=False)
+
+        # Compute the sum of W_i^T * rho_i^-2 * X_i, and the sum of traces
+        # of X_i^T * rho_i^-2 * X_i
+        wt_invpsi_x = np.zeros((self.features, samples))
+        trace_xt_invsigma2_x = 0.0
+        for subject in range(subjects):
+            if data[subject] is not None:
```
qihongl, line 616: referred in a comment above

qihongl (commented):
Hi, I'm not familiar with mpi but I think the code makes sense (in terms of matching the formulas in the paper)! 
I think some input validation for `data` at the beginning of this function would be useful here. And I have 2 other very minor comments. Ah, and a test would be nice. Perhaps a test that computes the score for some toy data? 

Thanks!

brainiak/funcalign/srm.py
qihongl: ah and this function probably want to assert `len(data) >= 2`

PR #480, opened 2020-09-05
Update ISC test documentation with caveat about FPRs

snastase:
I added caveats about the nonparametric ISC tests "not strictly controlling false positive rates" in consideration of #466. @cbaldassano—let me know if you any think this is sufficient, or if you having any suggestions about wording.


cbaldassano (approved):
Looks good to me!

Issue #481, opened 2020-09-30
fmrisim compute_signal_change


Matthew-Ryan1995:
Hey,

I have been working through the example for the use of fmrisim found at https://github.com/brainiak/brainiak/blob/master/examples/utils/fmrisim_multivariate_example.ipynb and trying to adapt this to my own example.   I think there might be an error in chunks 46 and 47 when computing `noise_func`.

As it stands, `noise_func` and `signal_func` have different shapes (27 x 294 and 294 x 27 respectively) which is not getting detected by the error check in `compute_signal_change`  (line 3185 in fmrisim.py).  Would I be right in saying that `noise_func` should be transposed so it is of the form **Time x Voxel** ?

I came across this when I was trying to place signal in an example where I had more voxels of interest than time points and I hit an indexing error.

Cheers,
Matt

CameronTEllis:
Hi Matt,

Great catch, thanks for finding this. There is a little bit of a quirk here that I will explain a bit but ultimately you are correct there is an error that needs to be fixed.

So the check I added on line 3185 of fmrisim would catch this, but the current version of BrainIAK that you can download using conda doesn't have that update to the code since it was a relatively recent addition. Hence, it slips through if you run this. 

However, these variables should be the same shape (they didn't use to be in an earlier versions of the code). Hence I have updated the notebook to suit in a new pull request.

PR #461, opened 2020-05-05
Unpin scikit-learn and set gamma to auto in SVC tests

ariekahn:
Fixes #448 

scikit-learn 0.22 changed the default value of `gamma` in `SVC` to 'scale' from 'auto' (see https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html). This PR should hopefulyl fix the tests using SVC by explicitly setting gamma to 'auto' in line with previous behavior.

PR #483, opened 2020-10-01
Update fmrisim example

CameronTEllis:
Added a table of contents and fixed a bug raised in #481.


mihaic:
@CameronTEllis, are you sure you want to remove the `gamma` parameter from the `SVC` call? We added it to maintain the existing behavior when using scikit-learn>=0.22:
https://github.com/brainiak/brainiak/pull/461/files#diff-ad1e5d506b6c5b8836bc3c0becdba1af

CameronTEllis:
@mihaic Thanks for pointing this out, I didn't know that this was the fix. I made that change

mihaic (approved):

PR #495, opened 2021-04-10
srm: subtract the subject-specific mean in srm.transform() and srm.transform_subject()

qihongl:
This is a continuation of  #406


qihongl:
I updated SRM to use the intercept term. Then in the srm test, I added an example showing that ... 

1. if I have a bunch of subjects in the form of 

X_i = W_i.T @ S  + intercept_i

Then SRM can align them (only possible with the intercept term)

2. After fitting, if I have a new subject in the form of 

X_new = W_new.T @ S  + intercept_new

Then srm.transform_subject() can align it to a trained shared response. It returns the estimated w and intercept (called `w` and `mu`). A short comment about how to use `w` and `mu` is added to the doc string. 

I also added comments in detsrm, sssrm, rsrm about they current don't use the intercept term.

qihongl:
Hi @mihaic I made a new PR about SRM intercept here.

TuKo:
I think that this is still a problem in the other SRM classes. 
Instead of documenting the differences, why don't we just do the respective modifications (at least in `DetSRM`).

TuKo:
Other than that, the change for `SRM` looks good to me!

qihongl:
Thanks! @TuKo But does `detSRM` estimate `mu` during model fitting? I thought it is just estimating n_units x n_units orthogonal transformations.

qihongl:
btw I think we should recommend people normalizing the data before using any SRM. Assuming the data is standardized, all SRMs can work well even without adjusting for subject-specific intercept terms. Of course, normalization also handles subject-specific scaling, which is also important. 

Fortunately, this seems to be what people have been doing.

TuKo:
> Thanks! @TuKo But does `detSRM` estimate `mu` during model fitting? I thought it is just estimating n_units x n_units orthogonal transformations.

No, it doesn't. But it should have that option, if we are going to assume that data is not z-scored.

qihongl:
I see. I just modified detsrm to allow it to handle intercepts. I also added a test to show that now detSRM can align subjects with different intercept terms (the same as the SRM intercept test).

qihongl:
There is another failed test that I don't understand. @TuKo do you know why it is failing?

mihaic:
The unit tests are passing. The documentation building is failing, but I can fix that, don't worry about it.
```
================ 199 passed, 1394 warnings in 527.82s (0:08:47) ================
```
https://travis-ci.org/github/brainiak/brainiak/jobs/767874844

qihongl:
Sounds great! Thanks! @mihaic

qihongl:
Hi @mihaic @TuKo it just occurred to me -- is there anything I can help with for this PR? Thanks!

Issue #499, opened 2021-04-15
dev: Switch CI from Travis to GitHub


mihaic:
Travis announced some time ago that travis-ci.org will stop working and now it finally happened. We can either move to travis-ci.com or somewhere else. I propose GitHub.

mihaic:
Fixed by PR #502.

Issue #501, opened 2021-05-24
mpiexec hangs with multiprocessing with a non-fork start method


mihaic:
See the following example:
```bash
mpiexec -n 2 python3 test_pool.py  # runs forever, kill it manually
cat test_pool.py
```

```python3
import multiprocessing

from mpi4py import MPI

def test_pool():
    multiprocessing.set_start_method("spawn")
    with multiprocessing.Pool(2) as pool:
        pool.apply(f, ())

def f():
    pass

if __name__ == "__main__":
    test_pool()
```

This affects us when running on macOS, where the the start method is "spawn". I will disable the affected tests so CI can pass.

We should probably disable multiprocessing in this situation.

PR #502, opened 2021-05-25
Add GitHub Actions

mihaic:
Fixes issue #499.

Also made a number of other changes in the process of getting GitHub
Actions to run. I think it makes sense to include them here as opposed
to making separate PRs for them.

Add workaround for editable user install.

Ignore flake8-docstrings errors (not part of CONTRIBUTING).

Use flake8 extend-ignore instead of explicit ignore.

Add IGNORE_CONDA env variable for pr-check to use Venv even when Conda
is available.

Drop unused Jenkins script.

Remove editable install from requirements-dev.txt to speed up testing.

Fix missing matnormal in pr-check.

Disable tests with mpiexec with non-fork multiprocessing (issue #501).


mihaic:
I think the GitHub Actions workflow will not run here. You can see the status on my fork for the first commit (06e3bac):
https://github.com/mihaic/brainiak/actions/runs/873056743

gdoubleyew (approved):

PR #511, opened 2022-02-21
Notebook testing

davidt0x:
This PR implements CI testing of brainiak and the example notebooks on Princeton research computing cluster della. I am still trying to work out some issues with failing or hanging tests on della. For now, I will have these tests skipped.

PR #526, opened 2023-02-26
change dtype=np.int to dtype=int

lcnature:
In [release 1.20 of numpy](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations), the use of `np.int` was depreciated. Users with numpy>1.19 in their virtual environments will encounter errors due to a few usage of np.int. Per the suggestion, a default way is to change to `dtype=int` from `dtype=np.int`, although what exact type of int will be used depends on the system. However, since most of the experiments won't have subject number more than 2^32 and I suppose it is unlikely that a system defaults to int8. I think leaving it to `dtype=int` should be fine

### DIFF
--- a/brainiak/funcalign/srm.py
+++ b/brainiak/funcalign/srm.py
@@ -240,8 +240,8 @@ class SRM(BaseEstimator, TransformerMixin):
                     "Not all ranks have same number of subjects")
 
         # Collect size information
-        shape0 = np.zeros((number_subjects,), dtype=np.int)
-        shape1 = np.zeros((number_subjects,), dtype=np.int)
+        shape0 = np.zeros((number_subjects,), dtype=int)
+        shape1 = np.zeros((number_subjects,), dtype=int)
 
         for subject in range(number_subjects):
             if X[subject] is not None:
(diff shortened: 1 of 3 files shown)


buildbot-princeton:

lcnature:
@mihaic can you review this? Thanks!

mihaic (approved):

lcnature:
It looks like some errors arose during the code checking, that some lines were too long. Shall we change the code to conform to the formatting requirement?

lcnature:
> @lcnature, let's work on the minor test issues after we merge the major changes in #511 which will greatly improve the PR workflow.

Thanks! So I suppose I just need to wait? Anything I can help? I am not super familiar with the testing workflow.

mihaic:
Right, nothing to do but wait. PR #511 is taking a while because it involves setting up testing on Princeton infrastructure.

adswa:
Hey :) I've been monitoring this PR for a while, because I had fought quite a few version incompatibilities for an analyses with a wide range of packages. I needed to eventually resort to creating my own fork with this change in order for my analysis to run. Now I would like to containerize my analysis, and it would be ideal if I could base it on the original brainiak repository rather then my custom fork. I was wondering if, given that #511 was merged, this PR could get renewed attention? 
Thanks for all the work on this library!

mihaic:
Sorry for the long wait, @adswa. Hopefully you can use the upstream BrainIAK code now.

PR #539, opened 2023-12-11
Update setuptools_scm

mihaic:


JiwaniZakir (commented):
The migration to `[tool.setuptools_scm]` with `version_file = "brainiak/_version.py"` is the correct approach for setuptools_scm>=8, which replaced the older `write_to` key. One thing worth verifying: if `setup.cfg` or `setup.py` still contains a `[tool:setuptools_scm]` section or a `use_scm_version` call with a conflicting `write_to` setting, the two configurations could conflict or produce duplicate/stale version files. The generated `brainiak/_version.py` should also be confirmed present in `.gitignore` — auto-generated version files should not be committed, and omitting that can cause confusing diffs. Additionally, bumping `setuptools>=62` is a meaningful constraint increase; it would be worth a note in the changelog or commit message explaining that this is driven by the setuptools_scm>=8 dependency requirement, so downstream packagers aren't surprised by the tighter floor.

PR #542, opened 2024-07-05
Burn in is removed so that the size of the signal and noise volumes are the same

CameronTEllis:
The example currently generates a signal volume that has a burn in it, but that is not added to the noise volume. That creates volumes of different sizes. We have removed the burn in to simplify things.


lcnature (approved):
I think this should fix the bug.

PR #549, opened 2024-11-15
Move to pyproject_toml + package cleanup

davidt0x:
This is a very large PR that began with a simple goal of moving the project over to pyproject TOML standard. Do to continued issues with failing CI jobs due to things like upper pins on old packages, dependencies on deprecated packages (theano), failing mpi tests, and other things; the PR became much more complicated. A rough list of improvements are the following:

- Moved to pyproject.toml standard
- Refactor package to generally recommended src layout
- Moved build to CMake + scikit-build-core
- Support for more recent pythons (3.12), 3.13 is limited by TF support now 
- Building wheels now on all major platforms (linux, macos ARM/x64, and windows)
- MPI tests are marked now and run in separate process using a slightly modified fork of https://github.com/minrk/pytest-mpiexec
- Removed almost all upper pins on packages, only TF probability remains.
- Conda packages building on linux and mac using mamba build, no windows here because conda-forge TF support on windows seems gone now
- Removed Theano as a dependency by porting ssrm to use tensforflow instead for differentiation. Tests pass but probably someone should look at this code more carefully to make sure nothing is broken.
- Fixed a large number of warnings all throughout the code
- Probably some other small things I am forgetting.

I still think there is considerable work to be done in cleaning up the pr-check.sh, run-tests.sh, and run-checks.sh scripts. Probably better to move these to github actions workflows or something so we can run them on windows. Currently, windows checks are just running pytest.


davidt0x:
Hey @mihaic , I think this one is ready, let me know what you think. della_notebooks run needs to be fixed. I will do another PR for it though.

PR #551, opened 2024-11-16
Trigger CD on tag pushes

mihaic:


davidt0x:
Hey @mihaic, the way I have it setup I think it should trigger if you make a release from github. I think this adds the tag as well.

mihaic:
OK, let's use GitHub releases.

PR #554, opened 2024-11-25
Fix anaconda release

davidt0x:
The anaconda upload command was just echoing and not actually running. This fix should hopefully work.


davidt0x:
I remove the della_notebooks workflow for now, it needs to be reworked anyway with new updates. I don't think it should take much now but I don't want to hold up the release.

mihaic:
I think it the issue was the merge queue. I disabled it. If we ever want to re-enable it, we should add `merge_group` as a workflow triggering event, otherwise required checks will not run to satisfy the queue:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions

PR #558, opened 2024-12-09
Document Windows support; fix Conda instructions

mihaic:


docs/installation.rst
```diff
@@ -2,27 +2,47 @@
...
+
+You can install BrainIAK wheels from PyPI on Linux, macOS, and Windows, e.g. using pip::
+
+    python3 -m pip install brainiak
+
+
+MPI
+---
+
+On Linux and macOS, there are no MPI4PY binaries in PyPI, so you need to install MPI separately, e.g., following the BrainIAK `source installation instructions <#requirements>`_.
```
davidt0x, line 20: You can get wheels for mpi4py on linux and macos (see [here](https://github.com/mpi4py/mpi4py-publish)). You need to add the extra index argument though. Some on linux, a one-line install looks like this:

`pip install brainiak[matnormal]==0.12.a5 impi-rt --extra-index-url https://pypi.anaconda.org/mpi4py/simple`

I just tested this in WSL and it worked for me, all tests passed. For windows, its even simpler since there are mpi4py wheels already on PyPI. Just this works:

`pip install brainiak[matnormal]==0.12.a5 impi-rt`

For MacOS, you can get the wheels similar to linux but I believe you still need to brew install or macports install opemmpi or mpich.

docs/installation.rst
```diff
@@ -2,27 +2,47 @@
...
 
 Source
 ++++++
 
 Requirements
 ============
 
-We support Linux and MacOS with Python version 3.5 or higher. Most of the
-dependencies will be installed automatically. However, a few need to be
+We support Linux, MacOS, and Windows with Python version 3.9 or higher. Most of
```
davidt0x, line 44: We don't support 3.13 at the moment, tensorflow is the problem I think.

davidt0x (approved):

PR #559, opened 2024-12-25
Fix Docker and add v0.12 release notes

mihaic:
Install from sdist in Dockerfile.

Update Matplotlib API in notebook example used to showcase Docker at brainiak.org.

Fix Docker instructions in README.

Include test data in sdist.


davidt0x (approved):
LGTM

